PR #603, opened 2022-06-24
602 only checkmate i2 @main

### LINKED ISSUE #602: Migration from `assertthat` to `checkmate`
This is the second incremental step from #597

Melkiades:
# Pull Request

Fixes #602


edelarua (approved):
LGTM @Melkiades!

shajoezhu (changes requested):
Thanks @Melkiades and @edelarua for looking into this. checks are failing. please update with

```
roxygen2::roxygenize('.', roclets = c('rd', 'collate', 'namespace'))
```

R/h_survival_duration_subgroups.R
```diff
@@ -336,14 +332,12 @@ h_coxph_subgroups_df <- function(variables,
...
-    is.character(variables$is_event),
-    is.character(variables$arm),
-    is.character(variables$subgroups) || is.null(variables$subgroups),
-    is.character(variables$strat) || is.null(variables$strat),
-    is_df_with_nlevels_factor(data, variable = variables$arm, n_levels = 2)
-  )
+  checkmate::assert_character(variables$tte)
+  checkmate::assert_character(variables$is_event)
+  checkmate::assert_character(variables$arm)
+  if (!is.null(variables$subgroups)) checkmate::assert_character(variables$subgroups)
```
gogonzo, line 338:
```suggestion
  checkmate::assert_character(variables$subgroups, null.ok = TRUE)
```

R/logistic_regression.R
```diff
@@ -373,7 +373,7 @@ h_simple_term_labels <- function(terms,
   assertthat::assert_that(
     is.table(table)
   )
-  assert_character_or_factor(terms)
+  checkmate::assertMultiClass(terms, classes = c("factor", "character"))
```
gogonzo, line 376: just to be consistent by using snake_case only
```suggestion
  checkmate::assert_multi_class(terms, classes = c("factor", "character"))
```

R/h_survival_duration_subgroups.R
Melkiades: thanks for this! I did not know you could do it this way. I will fix it where needed

Melkiades: I think this can be another issue for the future as there are many of these in the package and this PR is already huge (I will do it in the cases

R/logistic_regression.R
Melkiades: done in the next commit! thanks.

DESCRIPTION
```diff
@@ -23,7 +23,6 @@ Depends:
     R (>= 3.6),
     rtables (>= 0.5.1)
 Imports:
-    assertthat,
```
shajoezhu, line 26: 🥇

R/control_step.R
```diff
@@ -45,11 +45,11 @@ control_step <- function(biomarker = NULL,
                          num_points = 39L) {
   if (!is.null(biomarker)) checkmate::assert_numeric(biomarker)
   checkmate::assert_flag(use_percentile)
-  checkmate::assert_integerish(num_points, len = 1, lower = 2, any.missing = FALSE)
+  checkmate::assert_int(num_points, lower = 2)
```
shajoezhu, line 48: use `na.ok` to check missing?

R/summarize_variables.R
```diff
@@ -17,12 +17,10 @@
 control_summarize_vars <- function(conf_level = 0.95,
                                    quantiles = c(0.25, 0.75),
                                    quantile_type = 2) {
-  assertthat::assert_that(
-    all(vapply(quantiles, FUN = is_proportion, FUN.VALUE = TRUE)),
-    identical(length(quantiles), 2L),
-    is_proportion(conf_level),
-    quantile_type <= 9
-  )
+  checkmate::assert_true(length(quantiles) == 2)
```
shajoezhu, line 20: we have different implementation here for checking the length here

R/h_map_for_count_abnormal.R
```diff
@@ -92,10 +89,9 @@ h_map_for_count_abnormal <- function(df,
   df <- droplevels(df)
 
   normal_value <- setdiff(levels(df[[variables$anl]]), unlist(abnormal))
-  assertthat::assert_that(
-    # Based on the understanding of clinical data, there should only be one level of normal which is "NORMAL"
-    length(normal_value) == 1
-  )
+
+  # Based on the understanding of clinical data, there should only be one level of normal which is "NORMAL"
+  checkmate::assert_int(length(normal_value), lower = 1, upper = 1)
```
shajoezhu, line 94: we have a different implmentation using assert_true in this PR as well

R/h_step.R
```diff
@@ -72,15 +72,14 @@ h_step_trt_effect <- function(data,
...
+  checkmate::assert_multi_class(model, c("coxph", "glm"))
+  checkmate::assert_number(x)
+
   assert_df_with_variables(data, variables)
   arm_lvls <- levels(data[[variables$arm]])
-  assertthat::assert_that(
-    identical(length(arm_lvls), 2L)
-  )
+
+  checkmate::assert_int(length(arm_lvls), lower = 2, upper = 2)
```
shajoezhu, line 81: different implmentation in the same PR

R/logistic_regression.R
```diff
@@ -721,17 +700,17 @@ h_glm_inter_term_extract <- function(odds_ratio_var,
 #' @examples
 #' h_logistic_simple_terms("AGE", mod1)
 h_logistic_simple_terms <- function(x, fit_glm, conf_level = 0.95) {
-  assertthat::assert_that(inherits(fit_glm, c("glm", "clogit")))
+  checkmate::assert_multi_class(fit_glm, c("glm", "clogit"))
   if (inherits(fit_glm, "glm")) {
-    assertthat::assert_that(fit_glm$family$family == "binomial")
+    checkmate::assert_set_equal(fit_glm$family$family, "binomial")
```
shajoezhu, line 705: what about to use assert_true

_pkgdown.yml
```diff
@@ -320,15 +319,9 @@ reference:
       - to_n
 
   - title: Assertion Functions
-    desc: These functions supplement those in the assertthat package.
+    desc: These functions supplement those in the checkmate package.
     contents:
-      - all_elements_in_ref
-      - assert_df_with_factors
```
shajoezhu, line 326: keep this one?

NEWS.md
```diff
@@ -1,18 +1,33 @@
 # tern 0.7.8.9008
 
+
 ### Migration from `assertthat` to `checkmate`
-* `is_character_or_factor` is now `assert_character_or_factor` and uses 
-  `checkmate::assertMultiClass()` and made internal.
+* complete substitution of `assertthat` calls with `checkmate`.
+* `assert_df_with_factors`, `assert_equal_length`, and `assert_proportion_value`
```
shajoezhu, line 6: perhaps we should start a new util package to check these assertions, will be widely useful in other project as well. and we can export them

Melkiades: I think this would be perfectly fine in a utils' package. I would not create a package for assertions (it is only 200 lines) and it is not used downstream by any other package

R/logistic_regression.R
Melkiades: I adopted assert_set_equal because I think it is a bit more informative and it is possibly a better comparison method than `==`, but I do not know if it actually matters. It is possible to use assert_true ofc.

R/h_map_for_count_abnormal.R
Melkiades: yes, this is the most informative and I would keep this.

R/h_step.R
Melkiades: Similarly keeping this as standard for length checks

_pkgdown.yml
Melkiades: it is not used downstream nor in the catalogues, only internally

R/control_step.R
Melkiades: in assert_int the default for na.ok is FALSE

R/summarize_variables.R
Melkiades: I will apply the other implementation (next commit)

Melkiades: what do you think about these?
`checkmate::assert_true(length(strata_data) == nrow(df_tte))`

shajoezhu (approved):
lgtm! Many thanks @Melkiades

PR #604, opened 2022-06-24
docs: replaced github links with pkgdown

kpagacz:
Related to insightsengineering/coredev-tasks#301

# Pull Request

Fixes #nnn


nikolas-burkoff (approved):
Do we want to also change the link on line 1 (I think you did in teal?)

Otherwise all good

PR #610, opened 2022-06-28
393 Clean up pkgdown site documentation

### LINKED ISSUE #393: Incorrect reference list in the tern pkgdown
https://bookish-invention-97673167.pages.github.io/reference/index.html#statistics-functions
The one family functions (prefix with h_, a_, s_) are duplicated across different subsections, the structure is highly misleading. 
This comes from roxygen2 `describeIn` functionality as influence the pkgdown.yaml.

It looks like there is no control for that from the perspective of pkgdown.
https://github.com/r-lib/pkgdown/issues/986

One of the solutions is to rewrite the pkgdown yaml file to assume all siblings are in the one place.
The other solution is to rewrite the roxygen2 which does not looks to be a right one.
Or `man-roxygen` ...
Or simplify the reference and leave only the analyze functions (end product)

Ps. if updating pkgdown yaml please use `- starts_with("a_")` syntax to make code more elegant and easier to maintain.

edelarua:
Closes #393 

Completed to clean up the pkgdown site [Reference section](https://insightsengineering.github.io/tern/reference/index.html):
- [x] Updated `_pkgdown.yml` file to group functions and removed duplicate sections
- [x] General tidying of roxygen documentation in relevant files
- [x] Added external badge to `h_incidence_rate`, added internal badge to `create_afun_summary`, `create_afun_compare`, `prop_wilson`, `prop_clopper_pearson`, `prop_wald`, `prop_agresti_coull`, `prop_jeffreys`, `prop_diff_wald`, `prop_diff_ha`, `prop_diff_nc`, `prop_diff_cmh`
- [x] Split files with many functions into multiple: Added `tidy_cox.R`, `h_prop.R`, `h_prop_diff.R`, `h_prop_diff_test.R`, `h_coxreg.R`, `h_logistic.R` 
- [x] Removed duplicate footnotes functions

To build and view pkgdown site, run the following:
```r
pkgdown:::build_site_external()
```


R/compare_variables.R
```diff
@@ -353,20 +366,22 @@ a_compare.logical <- make_afun(
   .null_ref_cells = FALSE
 )
 
-#' @describeIn compare_variables Constructor function which creates a combined Formatted
-#'   Analysis function for use in layout creating functions [compare_vars()].
+#' Constructor Function for [compare_vars()]
+#'
+#' @description Constructor function which creates a combined Formatted Analysis function for use in
```
Melkiades, line 371: missing badge

R/compare_variables.R
```diff
@@ -353,20 +366,22 @@ a_compare.logical <- make_afun(
   .null_ref_cells = FALSE
 )
 
-#' @describeIn compare_variables Constructor function which creates a combined Formatted
-#'   Analysis function for use in layout creating functions [compare_vars()].
+#' Constructor Function for [compare_vars()]
```
Melkiades, line 369: If this is not described in the same document of compare_variables, it should be in a separate file. But as this is internal, Should it stay in that description doc? (I genuinely do not know this)

R/count_cumulative.R
```diff
@@ -1,15 +1,19 @@
...
 #' Summarize cumulative counts of a (`numeric`) vector that is less than, less or equal to,
 #' greater than, or greater or equal to user-specific thresholds.
 #'
 #' @name count_cumulative
-#'
 NULL
 
-#' @describeIn count_cumulative Helper function to calculate count and fraction of
-#'   `x` values in the lower or upper tail given a threshold.
+#' Helper Function for [s_count_cumulative()]
```
Melkiades, line 11: similar if not @ describeIn should have a separate file (also to match the documentation files).

Melkiades (changes requested):
In general great improvement! Just I think that many of the `@describeIn` should be reinstated as not doing so would put them in a separate doc file which in the case of internals is not reachable if not with `tern:::`

R/cox_regression.R
```diff
@@ -892,6 +255,232 @@ summarize_coxreg <- function(lyt,
   analyze_colvars(lyt = lyt, afun = afun_list[vars])
 }
 
+#' Fits for Cox Proportional Hazards Regression
```
Melkiades, line 258: I would split this file again in two because having two `#' @name` in one file can be very chaotic (you do not know where the source file is). So I would add a file called fit_coxreg with that (we can do this in another issue np)

R/g_step.R
```diff
@@ -113,17 +113,18 @@ g_step <- function(df,
   p
 }
 
-#' @describeIn g_step Custom Tidy Method for STEP Results
```
Melkiades, line 116: also here the `@describeIn` may be necessary

_pkgdown.yml
```diff
@@ -16,277 +16,113 @@ reference:
...
-      - s_cox_univariate
-      - s_length_proportion
-      - s_num_patients
-      - s_num_patients_content
-      - s_odds_ratio
-      - s_proportion
-      - s_proportion_diff
-      - s_summary
+  - title: Analysis Functions
+    desc: |
```
Melkiades, line 25: typo

edelarua:
Hi @shajoezhu, in the pkgdown Reference section functions are grouped by their .Rd file so I moved several to their own .Rd files to avoid grouping functions we don't want together.

Can you weigh in on whether we would should create separate .R files corresponding to each separate .Rd file or prefer to keep number of .R files to a minimum by grouping .Rd files of like functions within the same .R files (as was previously done with some h_ and d_ functions)?

Melkiades:
I think we may close this as it would be really too much work to merge with main atm. What do you think @edelarua?

edelarua:
> I think we may close this as it would be really too much work to merge with main atm. What do you think @edelarua?

I agree - it will be much easier to just create a new PR at this point. I'll close this PR.

PR #622, opened 2022-06-30
608 checkmate no stopifnot@main

### LINKED ISSUE #608: Replace all `stopifnot` with `checkmate`
`checkmate` is a better standard. More verbose and precise.

Melkiades:
# Pull Request

Fixes #608

### DIFF
--- a/R/decorate_grob.R
+++ b/R/decorate_grob.R
@@ -241,15 +241,29 @@ decorate_grob <- function(grob,
 #' @importFrom grid validDetails
 #' @export
 validDetails.decoratedGrob <- function(x) { # nolint
-  stopifnot(
-    grid::is.grob(x$grob) || is.null(x$grob),
-    is.character(x$titles),
-    is.character(x$footnotes),
-    is.character(x$page) || length(x$page) != 1,
-    grid::is.unit(x$outer_margins) || length(x$outer_margins) == 4,
-    grid::is.unit(x$margins) || length(x$margins) == 4,
-    grid::is.unit(x$padding) || length(x$padding) == 4
-  )
+
+  checkmate::assert_character(x$titles)
+  checkmate::assert_character(x$footnotes)
+
+  if (!is.null(x$grob)) {
+    checkmate::assert_true(grid::is.grob(x$grob))
+  }
+
+  if (length(x$page) == 1) {
+    checkmate::assert_character(x$page)
+  }
+
+  if (length(x$outer_margins != 4)) {
+    checkmate::assert_true(grid::is.unit(x$outer_margins) )
+  }
+
+  if (length(x$margins != 4)) {
+    checkmate::assert_true(grid::is.unit(x$margins) )
+  }
+
+  if (length(x$padding != 4)) {
+    checkmate::assert_true(grid::is.unit(x$padding) )
+  }
 
   x
 }
@@ -360,7 +374,8 @@ split_text_grob <- function(text,
     y <- grid::unit(y, default.units)
   }
 
-  stopifnot(grid::is.unit(width) && length(width) == 1)
+  checkmate::assert_true(grid::is.unit(width))
+  checkmate::assert_vector(width, len = 1)
 
   ## if it is a fixed unit then we do not need to recalculate when viewport resized
   if (!inherits(width, "unit.arithmetic") &&
@@ -388,11 +403,9 @@ split_text_grob <- function(text,
 #' @importFrom grid validDetails
 #' @export
 validDetails.dynamicSplitText <- function(x) { # nolint
-  stopifnot(
-    is.character(x$text),
-    grid::is.unit(x$width) && length(x$width) == 1
-  )
-
+  checkmate::assert_character(x$text)
+  checkmate::assert_true(grid::is.unit(x$width))
+  checkmate::assert_vector(x$width, len = 1)
   x
 }
 
--- a/R/decorate_grob.R
+++ b/R/decorate_grob.R
@@ -248,13 +248,18 @@ validDetails.decoratedGrob <- function(x) { # nolint
   if (!is.null(x$grob)) {
     checkmate::assert_true(grid::is.grob(x$grob))
   }
-
   if (length(x$page) == 1) {
     checkmate::assert_character(x$page)
   }
-  checkmate::assert_true(grid::is.unit(x$outer_margins))
-  checkmate::assert_true(grid::is.unit(x$margins))
-  checkmate::assert_true(grid::is.unit(x$padding))
+  if (!grid::is.unit(x$outer_margins)) {
+    checkmate::assert_vector(x$outer_margins, len = 4)
+  }
+  if (!grid::is.unit(x$margins)) {
+    checkmate::assert_vector(x$margins, len = 4)
+  }
+  if (!grid::is.unit(x$padding)) {
+    checkmate::assert_vector(x$padding, len = 4)
+  }
 
   x
 }
--- a/R/decorate_grob.R
+++ b/R/decorate_grob.R
@@ -252,18 +252,9 @@ validDetails.decoratedGrob <- function(x) { # nolint
   if (length(x$page) == 1) {
     checkmate::assert_character(x$page)
   }
-
-  if (length(x$outer_margins != 4)) {
-    checkmate::assert_true(grid::is.unit(x$outer_margins) )
-  }
-
-  if (length(x$margins != 4)) {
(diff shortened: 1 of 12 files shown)


R/decorate_grob.R
```diff
@@ -241,15 +241,20 @@ decorate_grob <- function(grob,
...
+  if (!is.null(x$grob)) {
+    checkmate::assert_true(grid::is.grob(x$grob))
+  }
+
+  if (length(x$page) == 1) {
+    checkmate::assert_character(x$page)
+  }
+  checkmate::assert_true(grid::is.unit(x$outer_margins))
+  checkmate::assert_true(grid::is.unit(x$margins))
+  checkmate::assert_true(grid::is.unit(x$padding))
```
edelarua, line 257: Can you incorporate the `|| length(x$y) == 4` part of these three assertions?

R/utils_checkmate.R
```diff
@@ -148,14 +150,16 @@ check_valid_factor <- function(x,
 }
 #' @describeIn assertions Check whether `x` is a valid factor (has levels and no empty string levels).
 #'   Note that `NULL` and `NA` elements are allowed.
-#' @param min.levels Minimum number of levels for `x`.
-#' @param max.levels Maximum number of levels for `x`.
-#' @param any.missing Default is `TRUE`, allowing missing values (`NA`).
+#' @param min.levels minimum number of levels for `x`.
+#' @param max.levels maximum number of levels for `x`.
+#' @param any.missing default is `TRUE`, allowing missing values (`NA`).
+#' @param ... additional parameters for [checkmate::assert_factor()]
```
edelarua, line 156: I think the standard here is to use `@inheritParams argument_convention` for common arguments like `...` (insert in documentation at beginning of file to use in all `assertions` functions)

R/decorate_grob.R
Melkiades: Good point! next commit fixes it

R/utils_checkmate.R
Melkiades: I do know if it is possible to include all parameters as I wanted to do

@inheritParams functionB, at least in Roxygen 6.0.0., will only inherit the parameters from functionA that also are found in functionB. Any parameters that are shared but that are defined in the functionA header will not be inherited from functionB

Also `...` is overwriting the other param on the same page... mmm I do not know exactly what to do here

edelarua: If you put proper descriptions for each parameter it should be fine either way. I just wasn't sure if you wanted to take advantage of the parameter inheritance here to simplify your code a bit.

Melkiades: Yes, I reason over it a bit and this is the cleaner way! Also, standards are identical, so no need to rewrite them. I reorganized it. Thanks!

R/utils_checkmate.R
```diff
@@ -77,12 +82,9 @@ check_df_with_variables <- function(df, variables, na_level = NULL) {
   }
   return(TRUE)
 }
-#' @describeIn assertions Check whether `df` is a data frame with the analysis `variables`.
+#' @describeIn Assertions Check whether `df` is a data frame with the analysis `variables`.
```
edelarua, line 85: Is there a reason why this was capitalized?

R/utils_checkmate.R
```diff
@@ -1,13 +1,18 @@
 #' Additional Assertions for `checkmate`
 #'
 #' @description
-#' We provide additional assertion functions which can be used together with [checkmate::assert()].
+#' We provide additional assertion functions which can be used together with [checkmate].
```
edelarua, line 4:
```suggestion
#' Additional assertion functions which can be used together with the `checkmate` package.
```

R/utils_checkmate.R
```diff
@@ -125,7 +127,11 @@ assert_df_with_variables <- checkmate::makeAssertionFunction(check_df_with_varia
 check_valid_factor <- function(x,
                                min.levels = 1,
                                max.levels = NULL,
```
edelarua, line 129:
```suggestion
                               min_levels = 1,
                               max_levels = NULL,
```

R/utils_checkmate.R
```diff
@@ -125,7 +127,11 @@ assert_df_with_variables <- checkmate::makeAssertionFunction(check_df_with_varia
...
-                               any.missing = TRUE) {
+                               empty.levels.ok = TRUE,
+                               null.ok = TRUE,
+                               any.missing = TRUE,
+                               n.levels = NULL,
```
edelarua, line 133:
```suggestion
                               empty_levels_ok = TRUE,
                               null_ok = TRUE,
                               any_missing = TRUE,
                               n_levels = NULL,
```
Fix linter errors by using snake_case instead of `.` in variable names

Melkiades: This was not causing the linter errors. I think it is the missing \cr in the docs.  We need these to have the dots to inherit parameters

Melkiades: No, you were right. Still doing that! I added `#nolint` so we can inherit parameters w/o rewriting them.

R/utils_checkmate.R
```diff
@@ -134,11 +140,12 @@ check_valid_factor <- function(x,
   }
   # main factor check
   res <- checkmate::check_factor(x,
-    empty.levels.ok = TRUE,
+    empty.levels.ok = empty.levels.ok,
```
shajoezhu, line 143: this is always true

R/utils_checkmate.R
```diff
@@ -125,7 +127,11 @@ assert_df_with_variables <- checkmate::makeAssertionFunction(check_df_with_varia
 check_valid_factor <- function(x,
                                min.levels = 1,
                                max.levels = NULL,
-                               any.missing = TRUE) {
+                               empty.levels.ok = TRUE, # nolint
```
shajoezhu, line 130: remove

R/utils_checkmate.R
```diff
@@ -177,7 +182,7 @@ check_df_with_factors <- function(df,
                                   variables,
                                   min.levels = 1,
                                   max.levels = NULL,
-                                  any.missing = TRUE,
+                                  any.missing = TRUE, # nolint
```
shajoezhu, line 185: @dinakar29 we were just confused here, sometimes the lintr fails with "." used in variable names, sometimes it doesn't. have you seen something like this happen before?

R/g_forest.R
```diff
@@ -154,19 +154,20 @@ g_forest <- function(tbl, # nolint
...
-    col_x > 0 && col_x <= nc,
-    col_ci > 0 && col_ci <= nc,
-    is.null(col_symbol_size) || col_symbol_size > 0 && col_symbol_size <= nc
-  )
+  checkmate::assert_number(col_x, lower = 0, upper = nc)
+  checkmate::assert_number(col_ci, lower = 0, upper = nc)
+  checkmate::assert_number(col_symbol_size, lower = 0, upper = nc, null.ok = TRUE)
+  checkmate::assert_true(col_x > 0)
+  checkmate::assert_true(col_ci > 0)
+  checkmate::assert_true(col_symbol_size > 0)
```
nikolas-burkoff, line 170: So a few lines above we have it is ok for `col_sumbol_size` to be NULL but if it is, this will fail:
image

R/g_forest.R
```diff
@@ -302,19 +303,18 @@ forest_grob <- function(tbl,
...
-    is.numeric(x) && length(x) == nr,
-    is.numeric(lower) && length(lower) == nr,
-    is.numeric(upper) && length(upper) == nr,
-    is.null(symbol_size) || length(symbol_size) == nr
-  )
+  if (is.null(vline)) {
+    checkmate::assert_true(is.null(forest_header))
+  } else {
+    checkmate::assert_number(vline)
+    checkmate::assert_character(forest_header, len = 2, null.ok = T)
```
nikolas-burkoff, line 311:
```suggestion
    checkmate::assert_character(forest_header, len = 2, null.ok = TRUE)
```

https://www.r-bloggers.com/2012/12/r-tip-avoid-using-t-and-f-as-synonyms-for-true-and-false/

Melkiades: that was an error sorry

R/g_forest.R
Melkiades: fixed! thanks

R/utils_checkmate.R
dinakar29: Nah that's new to me.
https://github.com/r-lib/lintr/issues/425 talks about something similar. The `# nolint` seems like the right solution.

PR #631, opened 2022-07-06
614 Update internal function examples to pass `--as-cran` checks

### LINKED ISSUE #614: Calls to `tern:::` in examples fails `--as-cran` because it throws a note
The problem is described in more detail [here](https://statisticaloddsandends.wordpress.com/2021/06/10/documentation-for-internal-functions/). 
The solution might be:

1. Move examples with internal functions in tests if there is not a duplicate already.
2. Use `\donotrun{}` flag and avoid using `:::` in the examples. Add the missing tests if there is not a duplicate already.

I am in favor of option 2 as it would allow the user to read and use the examples if necessary. Then it is necessary to have a way to understand if it is an internal function or not which can be done by a flag, comment, or starting the internal functions with dots.

edelarua:
Closes #614


DESCRIPTION
```diff
@@ -1,5 +1,5 @@
 Package: tern
-Title: Create Common TLGs used in Clinical Trials
+Title: Create Common TLGs Used in Clinical Trials
```
shajoezhu, line 2:
```suggestion
Title: Create Common TLGs Used In Clinical Trials
```

R/wrap_text.R
```diff
@@ -1,22 +1,28 @@
...
-#' teal.modules.clinical:::wrap_text(txt = text, width = grid::unit(4, "cm"), collapse = "\n")
-#' teal.modules.clinical:::wrap_text(txt = text, width = grid::unit(5, "cm"), collapse = "\n")
+#'
+#' # Internal function - wrap_text
+#' \dontrun{
+#' wrap_text(txt = text, width = grid::unit(4, "cm"), collapse = "\n")
+#' wrap_text(txt = text, width = grid::unit(5, "cm"), collapse = "\n")
+#' }
+#'
+#' @noRd
```
Melkiades, line 23: I might have added this, but I do not know, is it necessary?

_pkgdown.yml
```diff
@@ -333,6 +332,7 @@ reference:
       - fit_mmrm
       - starts_with("g_mmrm")
       - starts_with("h_mmrm")
+      - h_set_nest_theme
```
Melkiades, line 335: is this in the new nestcolor package?

Melkiades (approved):
great job here! 👍 It should be ok, but I was wondering if you checked that the tests have similar cases as the examples? Just because donotrun avoids running the examples and maybe we could lack a bit of coverage (but this can be another issue, I do not think there is unbalance between examples and tests).

R/wrap_text.R
edelarua: This was already here. I think this is a rarely used internal function so documentation was not required.

DESCRIPTION
edelarua: To pass `--as-cran` checks the title must use title case ("in" cannot be capitalized) or a warning is raised.

edelarua:
> > > great job here! 👍 It should be ok, but I was wondering if you checked that the tests have similar cases as the examples? Just because donotrun avoids running the examples and maybe we could lack a bit of coverage (but this can be another issue, I do not think there is unbalance between examples and tests).
> > 
> > 
> > Hi @Melkiades, we previously removed all `tern:::` instances from the examples, so after these changes there are no more occurrences of `:::` in the package. The only notes left when running `--as-cran` checks seem to be unrelated to this now.
> 
> yes, I just meant that it could happen that someone made a good example for the function but did not copy it in tests. I was wondering if you checked that (as far as I noticed, it is often not the case and examples and tests go together quite well)

There seemed to be tests for internal functions that might require them, but I did not explicitly check each function.

PR #639, opened 2022-07-14
generating log-rank p value by survdiff

### LINKED ISSUE #611: p value method for tte outputs requires refacoring
in SAS outputs, we usually have p value generated by log-rank method; however, in NEST, the default choice is `coxph_pairwise`(which gives log-rank statistic based on cox ph model, instead of the original log-rank test, see the NEST doc tte part)

The real corresponding testing should be done with `survdiff`

This can lead to minor discrepancies

raised by @clarkliming

JiaLiu0001:
close #611


shajoezhu (changes requested):
Thanks for the changes @JiaLiu0001 , could you add test case for this as well. Many thanks!

JiaLiu0001:
> Thanks for the changes @JiaLiu0001 , could you add test case for this as well. Many thanks!

Hi Joe, thank you for the comment. The test case for this issue is updated, please let me know if there's any further comment.

shajoezhu (approved):
Hi @JiaLiu0001 , very nice! thanks for the change. I am approving this PR now. Once you fix the spelling issue it is good to go. Thanks!

```
Run insightsengineering/r-spellcheck-action@v2
Run options(repos = c(CRAN = "https://cloud.r-project.org/"))
Loading required package: spelling
Run cd .
  WORD  FOUND IN
ph   NEWS.md:6[5](https://github.com/insightsengineering/tern/runs/7401668373?check_suite_focus=true#step:4:6)
Number of misspelled words: 1
You may correct the spellings of the words above or add them to the "inst/WORDLIST" file by running spelling::update_wordlist()
Error: Process completed with exit code 1.
```

R/survival_coxph_pairwise.R
```diff
@@ -100,9 +100,15 @@ s_coxph_pairwise <- function(df,
     ties = ties
   )
   sum_cox <- summary(cox_fit, conf.int = conf_level, extend = TRUE)
+  orginal_survdiff <- survival::survdiff(
+    formula_cox,
+    data = df_cox
+  )
+  log_rank_pvalue <- 1 - pchisq(orginal_survdiff$chisq, length(orginal_survdiff$n) - 1)
```
clarkliming, line 107: @JiaLiu0001 I think we should use `pchisq(lower.tail = FALSE)` instead of `1 - pchisq`. The former is accurate; although we don't have differences in printed out version (always <0.001 in both method)

PR #641, opened 2022-07-14
620 wald correction update@main

### LINKED ISSUE #620: Estimate_proportion_diff method Wald-cc not correct
in NEST, the Wald test for binomial difference is implemented using prop.test. This functions’ correction term is min(1/2*(1/n1+1/n2), abs(DELTA)) while delta is the observed proportion difference and n is the total number of patients in two groups respectively. According to the nest statistical book, the correction is called “Yate’s correction”

in stream, the method used is

proc freq 
 tables  /riskdiff(equiv correct type=wald)
if with correction, the correction term is always 1/2 *(1/n1+1/n2). The name of the correction is not specified in the SAS documentation, but some paper suggest that this is proposed by Yates.

In R, there is a equivalent function giving exact the same results, DescTools::BinomDiffCI(method="waldcc"). This function use a exact same correction term as SAS did.

Conclusion:
After looking into the reference papers, I believe this 1/2*(1/n1+1/n2) term is proposed by Yates and since we have identical solutions, then we should make NEST and stream align, otherwise we may face issues like “mismatches in SAS outputs and shiny apps”.
The “waldcc” method in NEST should be updated

reference:

Fleiss, J. L., Levin, B., and Paik, M. C. (2003). Statistical Methods for Rates and Proportions. 3rd ed. Hoboken, NJ: John Wiley & Sons.
Fagerland, Morten W., Stian Lydersen, and Petter Laake. “Recommended confidence intervals for two independent binomial proportions.” Statistical methods in medical research 24.2 (2015): 224-254.
Newcombe, Robert G. “Interval estimation for the difference between independent proportions: comparison of eleven methods.” Statistics in medicine 17.8 (1998): 873-890.

ayogasekaram:
# Pull Request

Replacing stats::proptest with DescTools equivalent BinomDiffCI for wald's test and correction.

Fixes #620


shajoezhu (changes requested):
Hi @ayogasekaram , many thanks for this change. Can you please add test case for this well. thanks!

R/prop_diff.R
```diff
@@ -96,6 +96,11 @@ prop_diff_wald <- function(rsp,
                            grp,
                            conf_level,
                            correct) {
+  if (correct == TRUE) {
```
pawelru, line 99: protip: `isTRUE()`

```
> x <- TRUE
> isTRUE(x)
[1] TRUE
> x == TRUE
[1] TRUE
> x <- iris
> isTRUE(x)
[1] FALSE
> x == TRUE
       Sepal.Length Sepal.Width Petal.Length Petal.Width Species
  [1,]        FALSE       FALSE        FALSE       FALSE   FALSE
  [2,]        FALSE       FALSE        FALSE       FALSE   FALSE
(...)
```

Melkiades (approved):
Lgtm! I was just wondering how big is `DescTools` as new dep?

shajoezhu:
> Lgtm! I was just wondering how big is `DescTools` as new dep?

DescTools is quite nice, and I think it will help us a lot, many things we can learn from the package, howeverI am inclined not to include DescTools as dependency into our package, as it is also going through development. some stats are right for some minor versions, but not in others. we need to do some investigation on this.

tests/testthat/test-table_rspt01.R
```diff
@@ -59,11 +59,11 @@ testthat::test_that("RSPT01: 1. Best Overall Response", {
       "60 (44.8%)", "(35.98, 53.57)", "40 (29.9%)", "(21.73, 37.97)",
       "9 (6.7%)", "(2.11, 11.33)", "24 (17.9%)", "(11.05, 24.78)",
       "1 (0.7%)", "(0.00, 2.58)", "ARM B", "(N=134)", "84 (62.7%)", "(54.1, 71.2)",
-      "", "-11.9", "(-23.7, -0.2)", "0.0416", "0.57 (0.34 - 0.96)",
+      "", "-11.9", "(0.2, 23.7)", "0.0416", "0.57 (0.34 - 0.96)",
```
shajoezhu, line 62: @ayogasekaram , can you check this please, seems this confidence interval have different sign, and the pointer estimator falls out side?

shajoezhu (changes requested):
Hi @ayogasekaram , just want to double check the sign's of the CIs. thanks

R/desctool_binom_diff.R
```diff
@@ -101,11 +100,11 @@ DescTools_Binom <- function(x1, n1, x2, n2, conf.level = 0.95, sides = c(
         CI.upper <- NA
       },
       score = {
-        w1 <- DescTools_BinomCI(
+        w1 <- DescTools::BinomCI(
```
shajoezhu, line 103: Hey @ayogasekaram , I was hoping we could use `DescTools_BinomCI` for this

shajoezhu (changes requested):
use `DescTools_BinomCI`

DESCRIPTION
```diff
@@ -28,6 +28,7 @@ Imports:
     checkmate,
     cowplot,
     dplyr,
+    DescTools,
```
shajoezhu, line 31: let's not introduce this package into the dependency,

shajoezhu (changes requested):
Hi Abi, many thanks for the changes. I have reviewed, and we will need to update the sign for the numbers. I will explain more in details via email.

R/prop_diff.R
```diff
@@ -89,36 +89,44 @@ d_proportion_diff <- function(conf_level,
...
-      correct = correct,
-      conf.level = conf_level
-    )$conf.int[1:2]
+    # check if binary response is coded as logical
+    checkmate::assert_logical(rsp, any.missing = FALSE)
+    checkmate::assert_factor(grp, len = length(rsp), any.missing = FALSE, n.levels = 2)
+
+    tbl <- table(grp, factor(rsp, levels = c(TRUE, FALSE)))
+    DescTools_Binom(
+      tbl[1], sum(tbl[1], tbl[3]),
```
shajoezhu, line 117: Might need to change the order here

shajoezhu (approved):
awesome! Thanks so much @ayogasekaram

shajoezhu:
Thanks so much @ayogasekaram . I have approved this PR, once the conflict is resolved. we can merge this back in!

shajoezhu:
Hi @ayogasekaram , the check is still failing, can you take another look please. Thanks!

cicdguy:
Ah shucks. This branch was called `620_wald_correction_update@main` instead of `620_wald_correction_update@pre-release`, and this is causing issues with the `tern` release with tests failing here: https://github.com/insightsengineering/tern/pull/682

@shajoezhu This is a release blocker. We need to get this resolved.

PR #643, opened 2022-07-19
589 Change `arm` and `extract` function names

### LINKED ISSUE #589: More specific naming for functions to avoid confusion, such as "arm" and "extract"
Checking the functions with @ayogasekaram, we noticed that some functions could be masking others from the env and viceversa. I noticed in particular the function `extract` that is an internal function of `tern` but could collide with `dplyr`. Similarly, I noticed that the (now) internal function `arm` is widely present as a variable name and widely mentioned in the documentation. If @shajoezhu agrees, I propose to use more specific naming for these two functions. 

Any suggestion or feedback is welcome!

PS: `arm`, when it is made internal fails the pkg-check, so I think this may hide a possible bug or a missing dependency in my script. I will explore this further.

- [x] arm
- [x] extract

edelarua:
Closes #589


Melkiades (approved):
Lgtm! We need just to be sure that also other packages that are using it are modified accordingly. My preliminary survey shows that no package downstream has a DIRECT call to those functions: it should be safe to proceed.

PR #679, opened 2022-08-10
still need to export combine_vectors because needed in tern.mmrm

danielinteractive:
part of https://github.com/insightsengineering/tern.mmrm/issues/13


shajoezhu (approved):
lgtm! Thanks @danielinteractive

PR #699, opened 2022-09-09
fix R CMD as-cran

nikolas-burkoff:
Part of https://github.com/insightsengineering/coredev-tasks/issues/334

After this PR we have only these (which are being taken care of later)
image

Fixed a url redirect, some line lengths in rd files and made a few examples run quicker so they are < 5 seconds on my machine now


edelarua (approved):
LGTM! Thanks @nikolas-burkoff

PR #704, opened 2022-09-13
Generalize data handling in tests

### LINKED ISSUE #703: Solution for data and library loading in `testthat`
At the moment data from `scda` is loaded multiple times (~9.9Mb of data each time) and libraries that are only suggested by `tern` are loaded multiple times at the beginning of each long test file. I strongly suggest following this [post](https://blog.r-hub.io/2020/11/18/testthat-utility-belt/) and the issue I filed in [rtables#393](https://github.com/Roche/rtables/issues/393). In other words, I think the best way to go here is to update the setup file in `testthat/` with the libraries needed across tests and add a helper file with data loading (always in `testthat/`) so they are both done once before tests are run. This would cut down test time and useless complexity and repetition across tests. Also, we need to consider if we should be using `withr`, but this is related to a later stage imo, when we have a more reasonable data caching design (as discussed [here](https://github.com/insightsengineering/scda/issues/98)). A tentative list of things to do here:

- [ ] Adding in setup libraries and polishing loads across tests.
- [ ] Create a helper file and centralize/polish the data used in tests (only `scda`-based ones).
- [ ] Discuss a new data design following [this](https://github.com/insightsengineering/scda/issues/98) and possibly `withr`.

What do you think @cicdguy @shajoezhu ?

Melkiades:
@nikolas-burkoff and @cicdguy, I am asking already your opinion as from here, it would be only incremental this PR while the basic concepts are here. It is still possible to do `withr::defer()` (for cleaning the env at the end [self-cleaning-test-fixtures](https://www.tidyverse.org/blog/2020/04/self-cleaning-test-fixtures/) and [test-fixitures vignette](https://testthat.r-lib.org/articles/test-fixtures.html)) and `memoise::memoise()` (to cache the load of variables, intro [here](https://memoise.r-lib.org/)) but it would involve micromanaging 1200 tests and I do not think it is worth it at the current state of things (i.e. with the current centralized solution). Let me know what you think, so I can continue this incremental task or modify it accordingly.

Fixes #703


tests/testthat/setup1-libraries.R
```diff
@@ -0,0 +1,8 @@
+# Libraries for tests (x are imports, xx are depends)
+library(scda)
+library(dplyr) # x
+library(rtables) # xx
+library(magrittr) # x
+library(broom) # x
+library(forcats) # x
+library(survival) # x
```
Melkiades, line 8: Regarding these, @cicdguy, I think they should be already loaded during tests, and, therefore, they could be taken out from tests, right? At the moment, there are 100+ library calls in `tern` tests that can be removed by this or removed completely.

nikolas-burkoff: yeah I think it's weird having library calls in tests - either the package is in suggests/imports (in which case use `pkg_name::func-name`) or it should already be loaded

tests/testthat/test-abnormal_by_marked.R
```diff
@@ -1,23 +1,17 @@
-library(scda)
-library(rtables)
-library(dplyr)
-
-adlb_raw <- local({
-  adlb <- synthetic_cdisc_data("rcd_2022_02_28")$adlb # nolintr
+# Local data pre-processing
+adlb_local <- local({
```
nikolas-burkoff, line 2: I wonder why we need local here...

nikolas-burkoff (commented):
Yeah I think this makes sense (esp if everything runs quicker now)

tests/testthat/test-abnormal_by_marked.R
Melkiades: Me too, is it actually freeing memory? I think all tests are in done in the same environment, so not really

tests/testthat/setup1-libraries.R
Melkiades: I just checked and only `library(scda)` is needed because it is in the suggested packages. Everything else is already imported in when doing tests. I will remove them

cicdguy: ^Concurring here. Thanks for cleaning these up.

tests/testthat/test-abnormal_by_marked.R
Melkiades: I am using it just when it is like a function and I do not need the intermediate steps to be in memory

tests/testthat/test-g_km.R
```diff
@@ -38,7 +33,7 @@ testthat::test_that("g_km default plot witch ci_ribbon = TRUE works", {
 })
 
 testthat::test_that("g_km plot with < = > in group labels works", {
-  df <- ex_adtte %>%
```
Melkiades, line 41: My question originates from this. This refers to `formatters`, which, I _re_ discovered, has the `random.cdisc.data` loaded in again, but there is no indication about its provenance. 

I know it is described as "used in tests only", but the last time this was updated is 8 months ago (see [here](https://github.com/insightsengineering/formatters/tree/main/data)). Also checking the `formatters` package, there is no use of `ex_*` data at all. I think it is used only 20-30 times in `rtables` tests, and it is worth discussing if it makes sense to have this outdated duplication of data. I suggest to replace all `ex_*` with `scda` well-maintained centralized data for testing. Let me know what you think about this @gmbecker and @nikolas-burkoff. 

I am convinced that is not convenient to use `ex_*` in `tern` at least and I am proceeding (causiously) to replace them.

Melkiades: As far as I remember our discussions @gmbecker, you want to keep `rtables` and `formatters` without an upstream dependency to `tern` so I guess the same is with `scda`. I understand but for the sake of generality and simplicity, it could be a good option to substitute `ex_*` in `rtables` with similar random data generated before the tests start

Melkiades:
The PR should be complete and the tests should run fine. @nikolas-burkoff, it should be ready for review. 

PS: time needed for the tests went from 256s to 136s on my machine ^^

cicdguy:
> The PR should be complete and the tests should run fine. @nikolas-burkoff, it should be ready for review.
> 
> PS: time needed for the tests went from 256s to 136s on my machine ^^

Excellently optimized! Thanks a bunch, @Melkiades!

tests/testthat/setup1-libraries.R
```diff
@@ -0,0 +1,3 @@
+# Extra libraries (suggested) for tests
```
nikolas-burkoff, line 1: So I prefer explicitly having dplyr:: and scda:: in the tests instead of loading the libraries, but I suspect that's too much of a change for this PR - though there seems to be lots of scda:: and dplyr:: calls so you might be able to remove them actually

tests/testthat/setup2-data.R
```diff
@@ -0,0 +1,54 @@
+# Data loading for tests
+test_data <- scda::synthetic_cdisc_data("rcd_2022_02_28")
+test_data_recent <- scda::synthetic_cdisc_data("rcd_2022_06_27")
```
nikolas-burkoff, line 3: probably wouldn't use "recent" -> could explicitly do test_data_2022_02 and test_data_2022_06

tests/testthat/setup2-data.R
```diff
@@ -0,0 +1,54 @@
...
+adcm_raw <- test_data$adcm
+advs_raw <- test_data$advs
+adeg_raw <- test_data$adeg
+adex_raw <- test_data$adex
+adlb_raw <- test_data$adlb
+admh_raw <- test_data$admh
+adrs_raw <- test_data_recent$adrs
+adtte_raw <- test_data_recent$adtte
+
+# Data as DM from `formatters` ## to discuss
```
nikolas-burkoff, line 22: to discuss...

tests/testthat/test-abnormal_by_worst_grade.R
```diff
@@ -1,23 +1,17 @@
...
-  adlb <- synthetic_cdisc_data("rcd_2022_02_28")$adlb # nolintr
-
+# Local data pre-processing
+adlb_local <- local({
   # Data set is modified in order to have some parameters with grades only in one direction
   # and simulate the real data.
-  adlb$ATOXGR[adlb$PARAMCD == "ALT" & adlb$ATOXGR %in% c("1", "2", "3", "4")] <- "-1"
-  adlb$ANRIND[adlb$PARAMCD == "ALT" & adlb$ANRIND == "HIGH"] <- "LOW"
-  adlb$WGRHIFL[adlb$PARAMCD == "ALT"] <- ""
+  adlb_raw$ATOXGR[adlb_raw$PARAMCD == "ALT" & adlb_raw$ATOXGR %in% c("1", "2", "3", "4")] <- "-1"
```
nikolas-burkoff, line 5: This is a really weird way of doing things. Having `adlb_local <- adlb_raw` then processing adlb_local would be much more usual but we can leave it here

tests/testthat/test-summarize_ancova.R
```diff
@@ -127,10 +125,9 @@ testthat::test_that("summarize_ancova works with healthy inputs", {
   testthat::expect_identical(result_matrix, expected_matrix)
 })
 
-
 testthat::test_that("summarize_ancova works with interaction", {
   iris_new <- iris %>%
-    mutate(p_group = case_when(
+    dplyr::mutate(p_group = case_when(
```
nikolas-burkoff, line 130:
```suggestion
    dplyr::mutate(p_group = dplyr::case_when(
```

tests/testthat/test-table_lbt05.R
```diff
@@ -1,8 +1,7 @@
-library(scda)
-library(dplyr)
+adsl <- adsl_raw
 
 get_adlb <- function() {
-  adlb <- synthetic_cdisc_data("rcd_2022_02_28")$adlb # nolintr
+  adlb <- adlb_raw # nolintr
```
nikolas-burkoff, line 4:
```suggestion
  adlb <- adlb_raw # nolint
```

we don't need the "r"

nikolas-burkoff (approved):
Looking good and much quicker (120s on my machine!) - a few comments in line and also note

image

tests/testthat/test-abnormal_by_worst_grade.R
Melkiades: yes, you are right. Good point. I think it is better to copy it locally and then reassign it so there is no modification to the general `*_raw` variable.

tests/testthat/setup1-libraries.R
Melkiades: I think it is always with `::`, but I would keep these two loads here still, at least until we need to remove the file (`setup-libraries.R`) entirely

PR #732, opened 2022-11-01
added legend to g_step plot

### LINKED ISSUE #22: Investigate how to add legend to `g_step()` plot
Maybe https://aosmith.rbind.io/2018/07/19/manual-legends-ggplot2/ can be helpful here, in particular:
image

sob2021:
closes #22

### DIFF
--- a/R/g_step.R
+++ b/R/g_step.R
@@ -108,12 +108,11 @@ g_step <- function(df,
     )
   }
   p <- p + ggplot2::geom_line(
-    ggplot2::aes_string(y = "y"),
-    color = est$col,
+    ggplot2::aes_string(y = "y", color = "est$col"),
     linetype = est$lty
   )
 
-  p <- p + ggplot2::labs(x = attrs$biomarker, y = attrs$estimate)
+  p <- p + ggplot2::labs(x = attrs$biomarker, y = attrs$estimate, color = "Legend")
   if (use_percentile) {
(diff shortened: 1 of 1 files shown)


danielinteractive (commented):
Thanks @sob2021 so I get from this the below plot.
Apologies that the description of the task was not specific enough - I guess I meant to say there that we would like to have the confidence level of the CI ribbon (the filled area) visible in the legend somehow e.g. as "95% CI", as well as maybe "Estimate" for the center line

image

danielinteractive:
ok idea is:
- use empty string or blank string as the legend title, since we will have two legends so that it does not look strange
- add `fill` as well same as `color` before
- use `scale_color_identity()` and `scale_fill_identity()` to define the correct labels

Melkiades:
This is the current state of things:
Screenshot 2022-11-08 122922
I did not use the identity functions but another way. If it is needed to add a different CI and Estimate label (to put as input), I can correct this

danielinteractive (approved):
thanks @sob2021 and @Melkiades , can you also add a short `NEWS` entry for this? afterwards good to merge from my side

Melkiades:
> @Melkiades awesome thanks for jumping in and helping here! Can you check the messages in the unit tests, somehow they are leading to failing checks here?

Thank you for the fast review! I do not understand very well why these errors. They do not happen locally. I will investigate further, maybe there is an invisible message :D

PR #739, opened 2022-11-09
401 add rate01 tlg@main

### LINKED ISSUE #538: Create summarize_glm_count.R from RATET01 design doc
# Motivation

This is part of adding the new template RATET01, see https://github.com/insightsengineering/tern/issues/319

# To-Do:

## Pull code out of design doc (tern/design/design_ratet01.Rmd) and put into separate file, summarize_glm_count.R
- [x] Helper functions: h_glm_quasipoisson, h_glm_poisson, h_glm_ppmeans, & h_glm_count
- [x] Statistics function: s_glm_count
- [x] Analysis function: a_glm_count
- [x] Summarize function: summarize_glm_count 
- [x] Add annotations in summarize_glm_count.R for documentation. See summarize_ancova.R for examples.
- [ ] Create unit tests

DO NOT include h_glm_negbin helper function for the first iteration of this implementation.

ayogasekaram:
added unit tests for summarize_glm_count

adding rate01 to tlg MR found here: https://code.roche.com/nest/docs/tlg-catalog/devel/-/merge_requests/162

closes #538


R/summarize_glm_count.R
```diff
@@ -0,0 +1,486 @@
...
+#'  in averaging predictions. Number of weights must equal the number of levels included in the covariates.
+#'  Weights option passed to emmeans function (hyperlink) (link to emmeans documentation)
+#'
+#' @examples
+#'
+#' library(scda)
+#' library(dplyr)
+#' anl <- synthetic_cdisc_dataset("latest", "adtte") %>%
+#'   filter(PARAMCD == "TNE")
+#' anl$AVAL_f <- as.factor(anl$AVAL)
```
shajoezhu, line 34: I don't think this part is needed, as the below example is not ran

R/summarize_glm_count.R
```diff
@@ -0,0 +1,486 @@
...
+    "rate" = "xx.xxxx",
+    "rate_ci" = "(xx.xxxx, xx.xxxx)",
+    "rate_ratio" = "xx.xxxx",
+    "rate_ratio_ci" = "(xx.xxxx, xx.xxxx)",
+    "pval" = "x.xxxx | (<0.0001)"
+  ),
+  .null_ref_cells = FALSE
+)
+
+#' @describeIn summarize_glm_count Layout creating function which can be be used for creating
```
shajoezhu, line 411: I think we can move from line 411 to the end of function summarize_glm_count to the top of this script, so the documentation is clearer. As all the other parts are internal, and examples won't need to be run.

tests/testthat/test-summarize_glm_count.R
```diff
@@ -0,0 +1,397 @@
+testthat::test_that("h_glm_poisson glm-fit works with healthy input", {
+  library(scda)
+  library(dplyr)
+  anl <- synthetic_cdisc_dataset("latest", "adtte") %>%
+    filter(PARAMCD == "TNE")
+  anl$AVAL_f <- as.factor(anl$AVAL)
```
shajoezhu, line 6: I think this part of data loading can be external to the testing functions. so the testing data can be reused, and will be more efficient.

shajoezhu (changes requested):
Thanks @ayogasekaram ! it looks really good. I made some samll change requests. thanks!

tests/testthat/test-summarize_glm_count.R
Melkiades: actually the libraries should be in setup_libraries and the dataset should be taken from the ones in setup_data

R/summarize_glm_count.R
Melkiades: I think it is needed for the last example which should be on the same doc file. Still, I would not repeat this there

tests/testthat/test-summarize_glm_count.R
shajoezhu: Excellent! That is the proper way to do it now! Thanks @Melkiades

R/summarize_glm_count.R
Melkiades: It makes sense but we can always put comment titles to describe the "internal" steps. Not sure. Maybe you are right and having only "runnable" docs is better

inst/WORDLIST
```diff
@@ -54,13 +54,16 @@ nd
 onlinelibrary
 paramcd
 pearson
+poisson
 pratt
 pre
+quasipoisson
+quasipossion
```
Melkiades, line 61: is quasipossion a thing? sorry, I may be very ignorant here but it seems a typo to me

Melkiades:
Great job Abi!! :) I did not review it in detail, as Joe is on it, but maybe we could aim at a bit higher coverage (>90%)

inst/WORDLIST
ayogasekaram: Good catch This was a spelling mistake in the R file. Thank you!

R/summarize_glm_count.R
shajoezhu: the last example has its own already

shajoezhu (approved):
changes lgtm! Thanks @ayogasekaram

PR #756, opened 2022-11-29
Complete pkct01

### LINKED ISSUE #277: PKCT01 to tern
image

Similar to PKPT03, but instead of split by param in PKPT03, split by visits, then split by sampling time

# Work

Each requires a pull-request 

## Summary Function
- [x] use `s_summary_numeric`

* `n`, `mean`,  `sd`, `cv_mean`,`geometric_mean`, `cv_geometric_mean` , `median`, `minimum`, `maximum`

Melkiades:
Fixes #277


R/argument_convention.R
```diff
@@ -39,6 +39,7 @@
 #' @param lyt (`layout`)\cr input layout where analyses will be added to.
 #' @param na.rm (`flag`)\cr whether `NA` values should be removed from `x` prior to analysis.
 #' @param na_level (`string`)\cr used to replace all `NA` or empty values in factors.
+#' @param na_str (`string`)\cr used to replace all `NA` with custom `string`.
```
edelarua, line 42: Is `na_str` the same as `na_level`? If so I would keep with the standard and use `na_level` instead.

edelarua (commented):
Looks great @Melkiades! Just one question about the parameter name. Can you update NEWS as well?

R/argument_convention.R
Melkiades: it makes sense. Done!

edelarua (approved):
Lgtm after those updates are in. Thanks @Melkiades!

shajoezhu (changes requested):
tests/testthat/test-table_pkpct01.R should be tests/testthat/test-table_pkct01.R

shajoezhu (commented):
lgtm! Thanks @Melkiades

shajoezhu (approved):
lgtm! Thanks @Melkiades

shajoezhu:
Hi @Melkiades , can you recompile the man docs

PR #766, opened 2022-12-09
rename summarize vars in cols

### LINKED ISSUE #765: Rename `summarize_vars_in_cols` with analyze and add some inputs (e.g. `nest = TRUE`)

Melkiades:
Fixes #765 

@shajoezhu I have a question here. I copied a parameter (`nested = TRUE`) that we need to fix ADAT03 for good. It is coming directly from `rtables`, therefore I inherited the parameter description from there. The only problem is that it does not have the same default standard as tern does. Imo it is fine like this, so we keep the original description, as it is eventually updated, but let me know if you think it is better to copy the description and fix its format.

Mmmh I see that this brings us also errors from `rtables`' docs:
```
WORD      FOUND IN
colvars   analyze_vars_in_cols.Rd:1[17](https://github.com/insightsengineering/tern/actions/runs/3656134276/jobs/6178229240#step:4:20)
rtables   analyze_vars_in_cols.Rd:117

Number of misspelled words: 2
You may correct the spellings of the words above or add them to the "inst/WORDLIST" file by running spelling::update_wordlist()
```

What is best to do? Calling in also @edelarua for an opinion

### DIFF
--- a/R/analyze_vars_in_cols.R
+++ b/R/analyze_vars_in_cols.R
@@ -79,28 +79,28 @@ NULL
 #' result <- build_table(lyt, df = adpp)
 #' result
 analyze_vars_in_cols <- function(lyt,
-                                   vars,
-                                   ...,
-                                   .stats = c(
-                                     "n",
-                                     "mean",
-                                     "sd",
-                                     "se",
-                                     "cv",
-                                     "geom_cv"
-                                   ),
-                                   .labels = c(
-                                     n = "n",
-                                     mean = "Mean",
-                                     sd = "SD",
-                                     se = "SE",
-                                     cv = "CV (%)",
-                                     geom_cv = "CV % Geometric Mean"
-                                   ),
-                                   labelstr = " ",
-                                   nested = TRUE,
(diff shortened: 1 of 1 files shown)


shajoezhu (changes requested):
Hi @Melkiades , thanks for the changes! It looks good. can you please update the NEWS as well. This has potentially quite a few downstream break changes for pk related tables. pkct01, pkpt03, pkpt05, and pkpt07.

NEWS.md
```diff
@@ -21,6 +21,8 @@
 ### Miscellaneous
 * Renamed `test_pkc01.R` and `summarize_variables_in_cols.R` into `test_pkct01.R`
   and `summarize_vars_in_cols.R`, respectively.
+* Renamed `summarize_vars_in_cols` into `analyze_vars_in_cols` to reflect the
```
Melkiades, line 24: Already added!

Melkiades: @shajoezhu

shajoezhu (approved):
lgtm! thanks @Melkiades

PR #787, opened 2022-12-28
expand score occurrences

### LINKED ISSUE #785: Modify and update `score_occurrences` so to have proper tests with different underlying `rtables` layouts to be sorted
- [x] modify original `score_occurrences` or add another score function that selects the columns (like `my_score_occurencies` does at the moment).
- [x] `test_score_occurences.R` uses summarize in the "broken-pagination" way should be fixed along with the previous point.

_Originally posted by @Melkiades in https://github.com/insightsengineering/tern/pull/776_

Melkiades:
Fixes #785


ayogasekaram (approved):
lgtm @Melkiades!

PR #790, opened 2023-01-09
fix summarize occurrences by grade for pagination

### LINKED ISSUE #783: Fix  `summarize_occurrences_by_grade` to work with pagination
>  `summarize_occurrences_by_grade` has the same problem of `summarize_num_patients` and needs to be fixed in a separate PR to fix `aet04`.
_Originally posted by @Melkiades in https://github.com/insightsengineering/tern/pull/776_

FYI

> aet03/04 has the same issue here
_Originally posted by @clarkliming in https://github.com/insightsengineering/chevron/issues/325#issuecomment-1319990325_

The latter is related to an older problem that can be found along with others in this thread: https://github.com/insightsengineering/chevron/issues/140
Please if it is possible also track down exactly the original comment where pagination is an issue for `aet03/aet04` @clarkliming. Feel free to add this issue to the one mentioning this if I have missed any. Thanks!

Melkiades:
It seems that the solution would be simply to use `count_occurrences_by_grade`, but I have not yet managed to make the sorting of the splitted values work.

Template to fix:

- [x] aet03
- [x] aet04

Fixes #783


edelarua (approved):
Lgtm @Melkiades!

PR #813, opened 2023-02-01
Update tests to use snapshots and use `tern` data

### LINKED ISSUE #779: update unittests that are using scda data

edelarua:
Closes #779


shajoezhu:
Thanks so much @edelarua ! very good start

tests/testthat/test-abnormal_by_worst_grade_worsen.R
```diff
@@ -83,17 +81,17 @@ testthat::test_that("h_adlb_worsen stacks data correctly", {
 
   expected_matrix <- structure(
```
Melkiades, line 82: Should we not start using snaps also here?

shajoezhu: let's do it in a seperate issue/PR

PR #814, opened 2023-02-02
checking if there will be an error with install-system-dependencies o…

arkadiuszbeer:
…n ubuntu 22 with old ver.of remotes

# Pull Request

Fixes #nnn


arkadiuszbeer:
Fix should be done in remotes pacakge.

PR #826, opened 2023-02-14
Deprecate footnotes and removing wrap_text

### LINKED ISSUE #736: Should we deprecate `footnotes()` function in tern?
I am wondering if we need this functionality, considering that it can be added as an attribute directly if needed internally and the fact that largely overlaps with `rtables`' footnotes

_Originally posted by @Melkiades in https://github.com/insightsengineering/tern/pull/734#discussion_r1012630118_

Melkiades:
Fixes #736


R/g_forest.R
```diff
@@ -246,12 +246,6 @@ g_forest <- function(tbl, # nolint
     vp = grid::plotViewport(margins = rep(1, 4))
   )
 
-  fn <- footnotes(tbl)
-  if (!is.null(fn)) {
-    footnotes(grob_forest) <- fn
-    message("grob footnote is not added to plot; suggest to use decorate_grob() to further decorate the grob")
-  }
-
```
Melkiades, line 254: @shajoezhu this is never used imo and it is bringing forwards little information

R/footnotes.R
```diff
@@ -12,14 +12,16 @@
 #' footnotes(x) <- "Species are equally distributed"
 #' attributes(x)
 `footnotes<-` <- function(x, value = NULL) { # nolint
+  dep_msg <- 'Please use current implementation in rtables or directly attr(x, "footnote")'
+  lifecycle::deprecate_warn("0.7.10", "footnotes()", details = dep_msg)
```
shajoezhu, line 16: @pawelru , I was wondering should this be 0.7.11?

shajoezhu (commented):
lgtm! Thanks @Melkiades. One minor question that I would like to check with Pawel with the version number, but otherwise everything is good. Thanks!

R/footnotes.R
Melkiades: I was wandering the same ahah

pawelru: Yes you are right - this should point to the future version - 0.7.10 is the current one

Melkiades:
@shajoezhu this is good to go ;)

shajoezhu (approved):
lgtm! Thanks @Melkiades

PR #843, opened 2023-03-02
Fix bug in `add_rowcounts`

### LINKED ISSUE #842: Fix bug in `add_rowcounts`
Currently `add_rowcounts` converts label rows to content rows with all values zero - this causes these rows/their child rows to be pruned whenever pruning is used on the table. This should be revised so that row count rows are only pruned when row counts are zero.

edelarua:
Closes #842 
Closes https://github.com/insightsengineering/chevron/issues/414

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -41,6 +41,7 @@
 * Fixed bugs in `rtables_access.R` that did not check for specific combination
   (also the standard values, that where never used) of column indices and names.
 * Fixed single applicable record bug in `count_abnormal_by_grade`.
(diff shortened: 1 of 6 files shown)


shajoezhu (commented):
thanks a lot @edelarua ! changes lgtm! I was wondering if you could quickly inspect the usage of this function in downstream scda.test, and chevron if this could cause any downstream break changes please

edelarua:
@shajoezhu no downstream changes in `chevron` or `scda.test`!

shajoezhu (approved):
lgtm! Thanks @edelarua

Issue #860, opened 2023-03-21
Implement `lubridate` with uniform timezones in `tern` data


edelarua:
### Feature description

Update date variables to use `lubridate` package, with the default "UTC" timezone for all dates. Regenerate datasets and update snapshots.

Date variables:
- ADSL: `TRTSDTM`, `TRTEDTM`, `EOSDY`
- ADAE: `ASTDTM`
- ADLB: `ADTM`
- ADRS: `ADTM`
- ADTTE: `ADTM`

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #863, opened 2023-03-22
Fix missing label for `TRTEDTM` in `tern` data [skip vbump]

edelarua:
Fixes bug in #860. All labels are now present.


ayogasekaram (approved):
lgtm @edelarua !

PR #876, opened 2023-04-05
858 missing values vignette

### LINKED ISSUE #858: Handling of missing values and `df_explicit_na` behaviour
Coming from https://github.com/insightsengineering/rtables/issues/579. This is particularly dear to me because the first times I had to work with `tern` I did not know of the existence of this function and wrote one on my own, which is a definition of reinventing the wheel. Surely, we need to have it a bit more connected to current documentation and/or a stand-alone vignette about pre-processing.

Nolan-Steed:
Fixes #858


Nolan-Steed:
I know in our previous conversation it was mentioned to make note of the response tables, but when I looked at the TLG-C for RSPT01, the pre-processing looked the same utilizing `df_expilict_na`. @shajoezhu can you remind me what I need to add in regards to that template? Thanks!

R/df_explicit_na.R
```diff
@@ -45,6 +45,102 @@
...
+#'
+#' result <- basic_table(show_colcounts = TRUE) %>%
+#'   split_cols_by(var = "ARM") %>%
+#'   add_overall_col("All Patients") %>%
+#'   summarize_vars(
+#'     vars = vars,
+#'     var_labels = var_labels
+#'   ) %>%
+#'   build_table(adsl)
+#' result
```
shajoezhu, line 143: Hi Nolan, thanks for the change. I think this part can go to a vignette document. These examples are very good, gives a bit more context. 

example here should be very light, something like

```
#' adsl <- tern_ex_adsl
#' adsl$AGE[adsl$AGE < 30] <- NA
#' adsl <- df_explicit_na(adsl)
```

what do you think?

R/df_explicit_na.R
```diff
@@ -15,9 +15,9 @@
 #' @param data (`data.frame`)\cr data set.
 #' @param omit_columns (`character`)\cr names of variables from `data` that should
 #'   not be modified by this function.
-#' @param char_as_factor (`flag`)\cr whether to convert character variables
+#' @param char_as_factor (`logical`)\cr whether to convert character variables
 #'   in `data` to factors.
-#' @param logical_as_factor (`flag`)\cr whether to convert logical variables
+#' @param logical_as_factor (`logical`)\cr whether to convert logical variables
 #'   in `data` to factors.
```
Melkiades, line 21: I think this was correct. I would revert these two.

see https://nest.pages.roche.com/docs/dev-guide/coding/documentation/

R/df_explicit_na.R
Melkiades: I agree with Joe. Maybe, I would keep a couple of practical examples showing different parameters

Melkiades (approved):
It seems good to me! Thanks Nolan :)

PR #878, opened 2023-04-06
Fix tibble snapshots

### LINKED ISSUE #866: Compare whole table for snapshot testing
I have noticed that for some of the snapshot tests (example [1](https://github.com/insightsengineering/tern/blob/main/tests/testthat/_snaps/abnormal_by_worst_grade_worsen.md?plain=1#L224-L248) [2](https://github.com/insightsengineering/tern/blob/main/tests/testthat/_snaps/abnormal_by_worst_grade_worsen.md?plain=1#L250-L274)) we are comparing `tibble` object using default print method that results in only small subset of table being printed out - on default it is just 10 rows. I feel it's not necessarily correct as a discrepancy might occur in e.g. row 11 and this won't be caught. We need to set [appropriate options](https://pillar.r-lib.org/reference/pillar_options.html) in the [testthat setup](https://testthat.r-lib.org/articles/test-fixtures.html) so as to compare all values. Alternatively we can use `expect_snapshot_value()` instead of `expect_snapshot()` if that's respect all values.

edelarua:
Converted most snapshot `tibble`s to `data.frame`s. Replaced 2 unnecessary `tibble` snapshots with different tests.

Closes #866

### DIFF
--- a/tests/testthat/_snaps/abnormal_by_worst_grade_worsen.md
+++ b/tests/testthat/_snaps/abnormal_by_worst_grade_worsen.md
@@ -3,47 +3,51 @@
     Code
       res
     Output
-      # A tibble: 20 x 8
-      # Groups:   USUBJID, PARAMCD [15]
-         USUBJID PARAMCD   VALUES     MIN   MAX WGRLOFL WGRHIFL GRADDR
-         <chr>   <chr>      <dbl>   <dbl> <dbl> <chr>   <chr>   <chr> 
-       1 4       ABC       0.0478  0.0478  187. "Y"     ""      Low   
-       2 3       OPQ       0.314   0.314   144. "Y"     ""      Low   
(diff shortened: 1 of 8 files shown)


Melkiades (approved):
Lgtm! Thanks Emily for this. I was just wondering how you selected tibbles to make them data.frame and do you think is it worth it to go the extra mile and add tests for col types? (imo not but maybe..)

edelarua:
I made all of the tibbles into data.frames except 2 which I chose to replace with other tests because they were huge tables and the function being tested was supposed to convert all values in the affected columns to the same value so it was much easier to just check that. 
I don't think that we need to test col types since only a few columns are affected by each function and we should (theoretically) be able to tell the output type from the function code itself, but if you think it's something we should add it doesn't hurt either way.

PR #902, opened 2023-05-10
deprecating `pairwise`

### LINKED ISSUE #899: Consider deprecating `pairwise`
The `pairwise` function is a duplicate of `univariate` and is seemingly unused (double check this) with no examples. If possible, we can deprecate the function.

Melkiades:
Fixes #899 

@danielinteractive fyi. If this function name is anyway needed we can make it an alias as it is identical to `univariate` in the source code


danielinteractive:
Thanks @Melkiades , I think this is just syntactic sugar, to make the code easier to read. Seems we are not using this directly in BAC, so should not impact BAC.

PR #909, opened 2023-05-10
Check all URLs

### LINKED ISSUE #885: All urls must resolve cleanly (without permanent redirects). This includes but is not limited to all github urls starting with `https` rather than `http`

edelarua:
Passing checks via `urlchecker::url_check()`.

Closes #885


Melkiades (approved):
Magnificent! We should do this for the other packages if they have some unprecise or broken links. Do we have something already going in this direction?

edelarua:
> Magnificent! We should do this for the other packages if they have some unprecise or broken links. Do we have something already going in this direction?

I don't think we've talked about this for other repos - we only brought this up for `tern` because we are putting it on CRAN, but I agree it would be nice to know that all the URLs are up-to-date and working since NEST users seem to have a lot of problems with broken links.

Issue #911, opened 2023-05-12
[Bug]: summarize_vars do not use `na_level`


clarkliming:
### What happened?

```
iris2 <- iris %>%
  mutate(Species = factor(Species, levels = c("setosa", "versicolor", "virginica", "asda")))
basic_table() %>%
  summarize_vars("Species", .stats = "count_fraction", na_level = "asda") %>%
  build_table(iris2)
```

since asda in in na_level, should it be excluded from output?

### sessionInfo()

```R
R version 4.2.2 Patched (2022-11-10 r83330)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.5 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] chevron_0.1.4.9010 checkmate_2.1.0    testthat_3.1.6    

loaded via a namespace (and not attached):
 [1] pkgload_1.3.2         tidyr_1.3.0           jsonlite_1.8.4       
 [4] splines_4.2.2         brio_1.1.3            shiny_1.7.4          
 [7] Rdpack_2.4            dm_1.0.5              yaml_2.3.7           
[10] remotes_2.4.2         sessioninfo_1.2.2     formatters_0.4.1.9001
[13] pillar_1.8.1          backports_1.4.1       lattice_0.20-45      
[16] rtables_0.6.0.3       glue_1.6.2            digest_0.6.31        
[19] promises_1.2.0.1      rbibutils_2.2.13      colorspace_2.1-0     
[22] htmltools_0.5.4       httpuv_1.6.9          Matrix_1.5-3         
[...]

PR #923, opened 2023-05-16
Fix na_level bug

edelarua:
# Pull Request

Fixes #nnn


edelarua:
> > Closes #911
> > I believe we should discuss refactoring this function entirely to process NA values in a more sensical way.
> 
> I think it is possible I added this. How do you think we should proceed?

I think we should remove the check for `NA` values - there's no reason why we shouldn't be able to process these in this/similar functions. Also, it wouldn't make a lot of sense for `na_level` to exist if we allowed `NA`s, since they would just show up in the table as `NA` and `na.rm` would still work - if the user wants to explicitly define/remove the `NA` level in pre-processing, they are still free to do so.

Melkiades:
> > > Closes #911
> > > I believe we should discuss refactoring this function entirely to process NA values in a more sensical way.
> > 
> > 
> > I think it is possible I added this. How do you think we should proceed?
> 
> I think we should remove the check for `NA` values - there's no reason why we shouldn't be able to process these in this/similar functions. Also, it wouldn't make a lot of sense for `na_level` to exist if we allowed `NA`s, since they would just show up in the table as `NA` and `na.rm` would still work - if the user wants to explicitly define/remove the `NA` level in pre-processing, they are still free to do so.

It makes sense. But what if we want just to replace output NAs with custom strings? I think that should remain, i.e. when we need to have "NE" instead. If there is another way than using `na_str`, I apologize for the silly question!

edelarua:
> > > > Closes #911
> > > > I believe we should discuss refactoring this function entirely to process NA values in a more sensical way.
> > > 
> > > 
> > > I think it is possible I added this. How do you think we should proceed?
> > 
> > 
> > I think we should remove the check for `NA` values - there's no reason why we shouldn't be able to process these in this/similar functions. Also, it wouldn't make a lot of sense for `na_level` to exist if we allowed `NA`s, since they would just show up in the table as `NA` and `na.rm` would still work - if the user wants to explicitly define/remove the `NA` level in pre-processing, they are still free to do so.
> 
> It makes sense. But what if we want just to replace output NAs with custom strings? I think that should remain, i.e. when we need to have "NE" instead. If there is another way than using `na_str`, I apologize for the silly question!

That does make sense and I think we should add that as well, but that's not the current function of the `na_level` argument in this function - currently, since `NA` is not allowed in the table, users can specify `na_level`, for example `na_level = "<Missing>"`, and this value will be treated as `NA` (i.e. removed by `na.rm`).

Melkiades (approved):
Lgtm! We can discuss later for the `na_str` propagation ;)

PR #924, opened 2023-05-17
Update NA handling by `s_summary`

### LINKED ISSUE #898: NA for s_summary
https://github.com/insightsengineering/tern/blob/2b31e0015f6d9059c5343bc88dfe44529270162c/R/summarize_variables.R#L124

Main question: when do we really need `na.rm=FALSE` for `s_summary`?

* for **numeric** values, if we don't exclude NA, then everything becomes NA. I think it is unrealistic to have it as an option (it should not be used)

* for **factors**, if we have NA in data, then `na.rm` does not work.  _(what if, I have some rows that should be excluded? should we make it NA before it enters into analysis, or we convert NA to some categories  and  remove it later ?)_
for factors, `na.rm = TRUE` remove the levels specified in `na_level`. but why bother to "create a manual level for NA and then remove that level", instead of using the `NA` directly (since in no cases should we use `na.rm = FALSE` for numerics, so for factor/character it is still true that NA should be excluded)

* for **characters**, similar to factors （but there is an error directly calling s_summary 
```s_summary(letters)
Error in checkmate::assert_string(x_name) : 
  argument "x_name" is missing, with no default
```
but the missing argument is actually `.var`. suggest to either refine the error message or not make .var as the required argument (I think the latter makes sense because anyway `s_summary` will not use that `.var`

edelarua:
Since `s_compare` is a wrapper for `s_summary` it has also been updated accordingly.

Updates:
- Allow `NA` values in input variables
  - For factor variables with `NA`s, if `na.rm = FALSE` an explicit `NA` level will be automatically added
  - If `na.rm = TRUE`, `<Missing>` will also be excluded - this way the `NA` behaviour will work correctly whether or not `df_explicit_na` has already been applied (with default settings)
- Change purpose of `na_level`
  - Old: string to consider as `NA` when setting `na.rm = TRUE`
  - New: string to print in lieu of `NA` values in output table

Closes #898


shajoezhu:
i wonder if we should leave this merge til the next PI.

edelarua:
> i wonder if we should leave this merge til the next PI.

This is fine to leave until next PI. It's not urgent since everything is working as is - just a nice improvement (in my opinion).

NEWS.md
```diff
@@ -15,6 +15,8 @@
 * Moved all KM plot documentation to the `g_km` function.
 * Added `a_count_patients_sum_exposure` for new analyze function `analyze_patients_exposure_in_cols` and `summarize_patients_exposure_in_cols`.
 * Added more informative error when the user selects an invalid method for unstratified analyses in `s_proportion_diff`.
+* Updated `s_summary` and `s_compare` to allow `NA` values in input variables. For factor variables with `NA`s, if `na.rm = FALSE` an explicit `NA` level will be automatically added. If `na.rm = TRUE`, `"<Missing>"` will also be excluded.
```
Melkiades, line 18:
```suggestion
* Updated `s_summary` and `s_compare` to allow `NA` values in input variables. For factor variables with `NA`s, if `na.rm = FALSE` an explicit `NA` level will be automatically added. `na.rm = TRUE` considers also `"<Missing>"` strings and excludes them.
```

NEWS.md
```diff
@@ -15,6 +15,8 @@
...
+* Updated purpose of `na_level` parameter in `s_summary` and `s_compare` to align with other `tern` functions: Previously used as string to consider as `NA` when setting `na.rm = TRUE`. Now used as string to print in place of `NA` values in output table.
```
Melkiades, line 19:
```suggestion
* Updated purpose of `na_level` parameter in `s_summary` and `s_compare` to align with other `tern` functions. Instead of being a string to consider as `NA` when setting `na.rm = TRUE`, it defines now a string to print in place of `NA` values in the output table.
```

R/summarize_variables.R
```diff
@@ -345,12 +345,15 @@ s_summary.factor <- function(x,
...
                              ...) {
-  assert_valid_factor(x, any.missing = FALSE)
+  assert_valid_factor(x)
   denom <- match.arg(denom)
 
-  if (na.rm) x <- fct_discard(x, na_level)
+  if (na.rm) {
+    x <- x[!is.na(x)] %>% fct_discard("<Missing>")
+  } else {
+    x <- x %>% explicit_na(label = "NA")
```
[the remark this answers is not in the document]
edelarua: Since we are generating a row for the NA level when `na.rm = FALSE` it needs to be explicitly defined. I figure that if people don't want the "NA" level they can call `df_explicit_na()` during pre-processing and set a custom string. 

If we leave missing values as `NA` it produces the following error: 
```
Error: Error applying analysis function (var - AVISIT): !anyNA(.names) is not TRUE
	occured at (row) path: root
```

edelarua:
> Also, have you checked if this affects `scda.test` in any way?

There are no templates/tests that use this argument as a default, so there is no change in `scda.test`, `tlg-catalog`, or `chevron` as a result of these changes.

Melkiades (approved):
LGTM! Great work Emily :)

PR #933, opened 2023-05-24
Clean up `.Rbuildignore`

### LINKED ISSUE #886: Root directory must be clean of non-standard files and directories (use `.Rbuildignore` to control this)

edelarua:
Closes #886


.Rbuildignore
```diff
@@ -4,16 +4,11 @@
...
 
-# design
-design/*
-
 # doc
-CODE_OF_CONDUCT.md
+^CODE_OF_CONDUCT\.md$
 ^docs$
 ^LICENSE\.md$
-README.*
```
shajoezhu, line 14: readme.md as well can be ignored?

Melkiades (approved):
Lgtm! Thanks Emily :)

.Rbuildignore
Melkiades: I do not think so. With this it should be tracked again correctly

edelarua: That's correct - `README.md` is a recognized file and does not need to be ignored.

PR #951, opened 2023-06-05
fix rmpt04 total level in `analyze_patients_exposure_in_cols`

### LINKED ISSUE #950: Feature requested  for having totals as another level in `analyze_patients_exposure_in_cols`
We need template `rmpt04` to have the totals as the last row. Doing this by modifying analyze wrapper function and adding a total flag for it.

Melkiades:
Fixes #950 

@BFalquet could you try this in your tests and see if it is enough to solve the problems? Thanks

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -30,7 +30,7 @@ jobs:
     name: R CMD Check 🧬
     uses: insightsengineering/r.pkg.template/.github/workflows/build-check-install.yaml@main
     secrets:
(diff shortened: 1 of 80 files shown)


BFalquet:
Works perfectly for me. thanks a lot.

edelarua (changes requested):
Just one suggestion for the param definition - otherwise it looks good!

edelarua (approved):
Thanks Davide!

PR #957, opened 2023-06-09
specify minimal version of dependencies

pawelru:
WIP


pawelru:
yay! min deps are fixed. max deps is to be resolved as most likely it would require changes in the codebase

PR #963, opened 2023-06-13
Fix `NA` value bug in `g_lineplot`

### LINKED ISSUE #962: [Bug]: `NA` values incorrectly removed in `g_lineplot`
Since `NA` values in the midpoint statistic are not removed, lines are not drawn if they include these values.

Bug reported on RocheOverflow.

edelarua:
Closes #962


Melkiades (approved):
Lgtm! Thanks Emily :)

PR #964, opened 2023-06-14
add `--as-cran` to local proj checks

### LINKED ISSUE #887: Package was checked locally with `--as-cran`, normal checking as the CI does is not sufficient

Melkiades:
I think this is the minimum that can be done. I am not sure if CI does check it already with `--as-cran` but, eventually, it is a fast fix, right @cicdguy?
Fixes #887


edelarua (approved):
Lgtm!

PR #970, opened 2023-06-15
update description and news version to 0.8.3, [skip vbump]

shajoezhu:
# Pull Request

Fixes #nnn

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: tern
 Title: Create Common TLGs Used in Clinical Trials
-Version: 0.8.2.9012
+Version: 0.8.3
 Date: 2023-06-14
 Authors@R: c(
     person("Joe", "Zhu", , "joe.zhu@roche.com", role = c("aut", "cre")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# tern 0.8.2.9012
+# tern 0.8.3
 
 ### Enhancements
 * Added explicit zero counts to `g_km` plot "at risk" annotation tables.


shajoezhu:
If there are references describing the methods in your package, please
add these in the description field of your DESCRIPTION file in the form
authors (year) <doi:...>
authors (year) <arXiv:...>
authors (year, ISBN:...)
or if those are not available: <https:...>
with no space after 'doi:', 'arXiv:', 'https:' and angle brackets for
auto-linking. (If you want to add a title as well please put it in
quotes: "Title")

- [x] We do not need "| file LICENSE" and the file as these are part of R.
This is only needed in case of attribution requirements or other
possible restrictions. Hence please omit it.

- [ ] "Using foo:::f instead of foo::f allows access to unexported objects.
This is generally not recommended, as the semantics of unexported
objects may be changed by the package author in routine maintenance."
Used ::: in documentation:
      man/cox_regression.Rd:
         tern:::a_coxreg(df = dta_bladder, labelstr = "Label 1",
variables = u1_variables, .spl_context = list(value = "COVAR1"), .stats
= "n", .formats = "xx")
      man/cox_regression.Rd:
         tern:::a_coxreg(df = dta_bladder, labelstr = "", variables =
u1_variables, .spl_context = list(value = "COVAR2"), .stats = "pval",
.formats = "xx.xxxx")
      man/summarize_patients_exposure_in_cols.Rd:
         tern:::a_count_patients_sum_exposure(df = df, var = "SEX",
.N_col = nrow(df), .stats = "n_patients")
Please omit one colon.

- [ ] You have examples for unexported functions. Please either omit these
examples or export these functions.

\dontrun{} should only be used if the example really cannot be executed
(e.g. because of missing additional software, missing API keys, ...) by
the user. That's why wrapping examples in \dontrun{} adds the comment
("# Not run:") as a warning for the user. Does not seem necessary.
Please replace \dontrun with \donttest.

Please unwrap the examples if they are executable in < 5 sec, or replace
dontrun{} with \donttest{}.

Melkiades (approved):
Lgtm! unblocking it if needed ;)

shajoezhu:
> Well, the point is that an examples suggests that a user could use it.
But unexported functions should not be used, hence non exporting but
ptoviding examples that show how to use this is somehow contradicting.

R/summarize_patients_exposure_in_cols.R
```diff
@@ -99,14 +88,14 @@ s_count_patients_sum_exposure <- function(df,
 #' * `a_count_patients_sum_exposure()` returns formatted [rtables::CellValue()].
 #'
 #' @examples
-#' tern:::a_count_patients_sum_exposure(
+#' a_count_patients_sum_exposure(
```
Melkiades, line 91: why this one is instead exported? Should be deleted as the other a_*

R/utils.R
```diff
@@ -9,19 +9,6 @@
...
-#' range_noinf(numeric(), na.rm = TRUE)
-#' range_noinf(c(1:5, NA, NA, Inf), na.rm = TRUE, finite = TRUE)
-#' range_noinf(Inf)
-#' range_noinf(Inf, na.rm = TRUE, finite = TRUE)
-#' range_noinf(c(Inf, NA), na.rm = FALSE, finite = TRUE)
-#' range_noinf(c(1, Inf, NA), na.rm = FALSE, finite = TRUE)
-#' }
-#'
 #' @keywords internal
 range_noinf <- function(x, na.rm = FALSE, finite = FALSE) { # nolint
```
Melkiades, line 13: Maybe these utils should be exported. Looks like useful

.Rbuildignore
```diff
@@ -7,6 +7,7 @@
 # doc
 ^CODE_OF_CONDUCT\.md$
 ^docs$
+LICENSE
 ^LICENSE\.md$
```
Melkiades, line 11: If not needed and could we just remove it from repo?

Melkiades (changes requested):
I think we need more than one reviewer here as there is a major removal of information. @edelarua @ayogasekaram what do you think?

edelarua (approved):
Lgtm! I agree with Davide's comments but overall it looks fine.

.Rbuildignore
shajoezhu: for the package installation, we won't need it. but as for the repo, I am inclined to keep it there

R/count_missed_doses.R
```diff
@@ -84,15 +73,6 @@ s_count_missed_doses <- function(x,
 #' @return
 #' * `a_count_missed_doses()` returns the corresponding list with formatted [rtables::CellValue()].
 #'
-#' @examples
-#' # Internal function - a_count_missed_doses
-#' \dontrun{
-#' #  We need to ungroup `count_fraction` first so that the `rtables` formatting
-#' # function `format_count_fraction()` can be applied correctly.
```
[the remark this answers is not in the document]
shajoezhu: i m removing examples for all internal functions for now, per reply for cran submission. 

let's reinstate these examples in the tern.s package as this is part of the PI

R/utils.R
shajoezhu: I think this can be migrated into the "tern.s" package as a statistics, would you agree?

Melkiades: maybe this one yes. The other ones in `utils.R` should be exported probably

shajoezhu:
hi @Melkiades , the package has been successfuly submitted. Can you unblock this merge, and we will address these as new issues. Thanks

Melkiades (approved):

shajoezhu:
awesome! Thanks guys!

PR #987, opened 2023-06-26
Fix tests - deprecate `pairwise` and add another test for `format_count_fraction`

### LINKED ISSUE #984: Add new tests
Output from the internal autovalidate report as of issue creation date:

```
Consistency checks failed:
 - requirement #0.073 'Formatting Count and Fraction' is not covered by any test
 - requirement #0.130 'Pairwise Formula Special Term' is not covered by any test
```

edelarua:
I'm not sure why `'Formatting Count and Fraction' is not covered by any test` was appearing since that function is covered by 2 tests already, but I've added another one in so that all lines in the function are tested just in case.

Closes #984


shajoezhu (approved):
Thanks @edelarua ! I flagged this before as a false positive before. hence we past last autovalidation

PR #998, opened 2023-07-05
Improve error message in `g_km` for single arm when annot_coxph=TRUE

### LINKED ISSUE #997: [Bug]: cryptic error message in tern::g_km
### What happened?

Reproducible example:

```
> library(tern)
> library(dplyr)
> library(ggplot2)
> library(survival)
> library(grid)
> library(nestcolor)
> 
> df <- tern_ex_adtte %>%
+   filter(PARAMCD == "OS") %>%
+   mutate(is_event = CNSR == 0)
> variables <- list(tte = "AVAL", is_event = "is_event", arm = "ARMCD")
> 
> df2 <- df[df$ARMCD == "ARM A", ]
> 
> res <- g_km(df = df2, variables = variables, annot_coxph = TRUE)
Error in grid::unit(as.numeric(height)/4, grid::unitType(height)) : 
  'x' and 'units' must have length > 0
```

It makes sense that it fails since if there is only one group; we cannot compute a hazard ratio, but the error message is not very helpful, would you consider adding a little assertion in the function? Something that throws an error message if there is only one value in the `arm` variable and `annot_coxph = TRUE`.

thank you

### sessionInfo()

_No response_

### Relevant log output

_No response_

### Code of Conduct

### Contribution Guidelines

### Security Policy

edelarua:
Closes #997

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,15 +2,14 @@
 
 ### Enhancements
 * Added method for `character` class to `h_coxreg_inter_effect` enabling `character` covariates in `summarize_coxreg`.
-
-### Enhancements
 * Started deprecation cycle for `summarize_vars` and `control_summarize_vars`. Renamed into `analyze_vars` and `control_analyze_vars` to reflect underlying `rtables` machinery while keeping backward compatibility with aliases.
+* Added assertion to `g_km` which checks whether there is only one arm present in the data when `annot_coxph` is true.
 
 ### Miscellaneous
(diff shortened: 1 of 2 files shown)


shajoezhu (commented):
hi @edelarua , thanks for the change! I was wondering if we can improve the assertion message a bit more intuitive, something like, it requires more than one arm if requires an annotation table for coxph

shajoezhu (approved):
lgtm! Thanks @edelarua 

there is a conflict in the news, can you update. thanks

PR #1017, opened 2023-08-01
Fix degenerate table warning in `summarize_coxreg`

### LINKED ISSUE #1016: Fix degenerate table warning when using `summarize_coxreg`
@clarkliming it is working correctly. The purpose of `assert_valid_table` is, as its name suggests, to check whether a table has a valid structure, or put another way, to assert that a table does not have a degenerate structure. 

The purpose is not to assert whether a particular table happens to be able to be rendered (ie, whether the undefined behavior present for degenerate table structures happens to fortuitously align so the table can be printed without error). 

In the case of this particular table, we have:

```
> assert_valid_table(result)
Error in assert_valid_table(result) : 
  Invalid table - found 1 (sub)structures which contain no data rows.
	The first occured at path:  root[STUDYID]->AB12345
  Use sanitize_table_struct() to fix these issues
```

The error tells us there is a substructure (ie subtable) in our table that does not have any data rows.  Using `table_structure`, we can see that this is in fact, correct:

```
> table_structure(result)
[TableTree] root
 [TableTree] STUDYID
  [TableTree] AB12345 [cont: 1 x 4] [TableTree] multivars
  [TableTree] AGE
   [ElementaryTable] ac_STUDYID_STUDYID_STUDYID_STUDYID (1 x 4)
  [TableTree] SEX
   [ElementaryTable] ac_STUDYID_STUDYID_STUDYID_STUDYID (1 x 4)
  [TableTree] RACE
   [ElementaryTable] ac_STUDYID_STUDYID_STUDYID_STUDYID (1 x 4)
```

This structure is weird, an

There are two ways to make this table non-degenerate with the same basic rendering:

1. Make that first row a content row on the root table, and then nest the split and analyze underneath it. That will have the result of repeating it in the case of horizontal pagination.
2. keep the studyid facet in the multivars row split, but use analyze instead of `summarize_row_groups` to implement it. In this case it will not be repeated across pagination.

Both approaches are achievable via narrow, straightforward changes to tern's `summarize_coxreg` function. cc @edelarua @Melkiades @anajens

_Originally posted by @gmbecker in https://github.com/insightsengineering/rtables/issues/701#issuecomment-1658942886_

edelarua:
Closes #1016 and https://github.com/insightsengineering/rtables/issues/701#

### DIFF
--- a/R/summarize_coxreg.R
+++ b/R/summarize_coxreg.R
@@ -370,16 +370,26 @@ summarize_coxreg <- function(lyt,
         common_var,
         split_label = "Treatment:",
         label_pos = "visible",
+        child_labels = "hidden",
         section_div = head(.section_div, 1)
-      ) %>%
-      summarize_row_groups(
-        cfun = a_coxreg,
-        extra_args = list(
-          variables = variables, control = control, multivar = multivar, eff = TRUE, var_main = multivar
-        )
       )
-    if (multivar) { # treatment level effects
+    if (!multivar) {
       lyt <- lyt %>%
+        analyze_colvars(
+          afun = a_coxreg,
+          extra_args = list(
+            variables = variables, control = control, multivar = multivar, eff = TRUE, var_main = multivar,
+            labelstr = ""
+          )
+        )
+    } else { # treatment level effects
+      lyt <- lyt %>%
+        summarize_row_groups(
+          cfun = a_coxreg,
+          extra_args = list(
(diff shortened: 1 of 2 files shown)


Melkiades (approved):
Lgtm! Thanks Emily :) I would only add a regression test with the same setting as the issue and maybe only check this with assert table

Melkiades (approved):

PR #1021, opened 2023-08-02
Update default `g_km` y limits

### LINKED ISSUE #1020: Default y limits of `g_km` plots should be dynamically set depending on data
### Feature description

Currently the default y limits for `g_km` plots are 0 and 1, which results in a lot of white space in some cases (e.g, when a maximum x-value is set). Default values should be set dynamically depending on the y-values of the data being plotted.

### Code of Conduct

### Contribution Guidelines

### Security Policy

edelarua:
Closes #1020

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,7 +54,8 @@ Suggests:
     nestcolor (>= 0.1.1),
     rmarkdown,
     stringr,
-    testthat (>= 3.0)
+    testthat (>= 3.0),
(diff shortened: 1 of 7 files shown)


Melkiades (approved):
Lgtm! thanks Emily. Could you add a small regression test maybe? I think at a certain point we need to considering testing plots here a bit better as it is done now in TLG-catalog (fyi @vedhav @pawelru)

PR #1023, opened 2023-08-04
Removal of internal `df_explicit_na`

### LINKED ISSUE #1019: [Question]: why h_adsl_adlb_merge_using_worst_flag calls `df_explicit_missing` inside the function
### What is your question?

`df_explicit_missing` is a dangerous step that affect the whole data. It should be executed by the user, like here 
image

however when in 
image
if the h_adsl_adlb_merge_using_worst_flag also contain the `df_explicit_missing` it can be too much for this function

reference: https://insightsengineering.github.io/tlg-catalog/tables/lab-results/lbt14.html

### Code of Conduct

### Contribution Guidelines

### Security Policy

Melkiades:
Fixes #1019

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 ### Miscellaneous
 * Fix swapped descriptions for the `.N_row` and `.N_col` parameters.
 * Fix bug in `analyze_vars_in_cols` when categorical data was used.
+* Removed internal `df_explicit_na`. It depends on the needs of users.
 
 # tern 0.8.5
 
--- a/R/h_adsl_adlb_merge_using_worst_flag.R
+++ b/R/h_adsl_adlb_merge_using_worst_flag.R
@@ -137,7 +137,6 @@ h_adsl_adlb_merge_using_worst_flag <- function(adsl, # nolint
   adlb_out$ATOXGR <- as.factor(adlb_out$ATOXGR)
   adlb_out$BTOXGR <- as.factor(adlb_out$BTOXGR)
 
-  adlb_out <- df_explicit_na(adlb_out)
   formatters::var_labels(adlb_out) <- adlb_var_labels
 
   adlb_out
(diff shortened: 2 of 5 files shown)


NEWS.md
```diff
@@ -3,6 +3,7 @@
 ### Miscellaneous
 * Fix swapped descriptions for the `.N_row` and `.N_col` parameters.
 * Fix bug in `analyze_vars_in_cols` when categorical data was used.
+* Removed internal `df_explicit_na`. It depends on the needs of users.
```
pawelru, line 6: I will understand this as removal of the function but you actually removed usage of it inside some of the functions

Melkiades, line 6:
```suggestion
* Removal of internal calls to `df_explicit_na`. Changes in `NA` values should happen externally to `tern` functions, depending on the users' needs.
```

NEWS.md
```diff
@@ -3,6 +3,7 @@
 ### Miscellaneous
 * Fix swapped descriptions for the `.N_row` and `.N_col` parameters.
 * Fix bug in `analyze_vars_in_cols` when categorical data was used.
+* Removal of internal calls to `df_explicit_na`. Changes in `NA` values should happen externally to `tern` functions, depending on the users' needs.
```
Melkiades, line 6:
```suggestion
* Removal of internal calls to `df_explicit_na`. Changes in `NA` values should happen externally to `tern` functions, depending on users' needs.
```

edelarua (changes requested):
Hi @Melkiades,

Could you add in a note so the user knows that `df_explicit_na` should be used during pre-processing? Specifically for `h_stack_by_baskets` the user should be aware that `na_level` needs to match the `na_level` used in `df_explicit_na`, otherwise the argument does nothing.

Thanks!

edelarua (approved):
Approving to unblock since I will be OOO next week.

shajoezhu:
Beniot also checked downstream. all is well. We are good to go! Thanks @Melkiades and @edelarua

PR #1032, opened 2023-08-16
Fix alias export for summarize_vars

Melkiades:


pawelru:
Shall we use lifecycle deprecation message?

shajoezhu (approved):
lgtm! thanks @Melkiades

Melkiades:
> Shall we use lifecycle deprecation message?

Done now. I added the current version, so we will have to wait. Note @BFalquet that using summarize_vars will generate a warning now

R/analyze_variables.R
```diff
@@ -715,5 +716,36 @@ analyze_vars <- function(lyt,
     section_div = section_div
   )
 }
-
-summarize_vars <- analyze_vars
+#' @inherit analyze_vars
```
pawelru, line 719: I think that a simple `@rdname analyze_vars` would be enough. Even without `@aliases`.

R/analyze_variables.R
```diff
@@ -715,5 +716,36 @@ analyze_vars <- function(lyt,
...
+summarize_vars <- function(lyt,
```
pawelru, line 720: out of curiosity:
can you do a simple:
```
summarize_vars <- function(...) {
  lifecycle stuff
  analyze_vars(...)
}
```

R/analyze_variables.R
Melkiades: Thanks, was looking for this

pawelru:
Please add one more entry in the `_pkgdown.yaml` in the deprecated functions section

R/analyze_variables.R
Melkiades: Warning: Skipping 'analyze_vars.Rd'
ℹ File lacks name and/or title

Melkiades: yep wrong name of the rd (analyze_variables). This probably need to change with the main function name

R/analyze_variables.R
Melkiades: done, much better
```r

## Select and format statistics output.
l <- basic_table() %>%
    split_cols_by(var = "ARM") %>%
    split_rows_by(var = "AVISIT") %>%
    summarize_vars(
        vars = "AVAL",
        .stats = c("n", "mean_sd", "quantiles"),
        .formats = c("mean_sd" = "xx.x, xx.x"),
        .labels = c(n = "n", mean_sd = "Mean, SD", quantiles = c("Q1 - Q3"))
    )

build_table(l, df = dta_test)
```

R/analyze_variables.R
Melkiades: I prefer describeIn so to add the deprecation badge

pawelru (approved):
good 👍  now you know how to deprecate functions :)

Issue #543, opened 2022-05-11
Implement 1/3 and 1/2 Imputation Rule in PKCT01 


qiliu1013:
![image](https://user-images.githubusercontent.com/24441440/167751761-28bc5176-0286-4617-acd0-c036d6cedfa3.png)

qiliu1013:
@shajoezhu, I went through the code and re-think about the solution we discussed yesterday.

```s_summary.numeric ``` is to calculate the stats in ADPC$AVAL. We cannot pre-process AVAL (ie assign -100 to BLQ records), it will change the stats of AVAL. We may have to bring another variable  into the condition with ```s_summary.numeric ``` or set up a separate functions for PKCT01. What do you think?

**Comment by shajoezhu** [comment]
hi @qiliu1013 , yes, you are correct.

what if we change the values to NA instead of -100? then the stats of AVAL can be correctly obtained, right?

alternative approach for setting up a seperate function is certainly possible, but will require some design

**Comment by qiliu1013** [comment]
@shajoezhu 
```
ADPC$AVALCAT1 <- ADPC$AVALC
ADPC$NFRLT <- ADPC$NRELTM1

summary_stats_pre <- function(data) {
    ltr_blq_ratio <- length(data$AVALCAT1 %in% c('BLQ','LTR','<PCLLOQ')) / nrow(data)

    if (ltr_blq_ratio <= 1/3) {
        list(
            Mean = mean(data$AVAL, na.rm = TRUE),
            Median = median(data$AVAL, na.rm = TRUE),
            Maximum = max(data$AVAL, na.rm = TRUE),
            Min = min(data$AVAL, na.rm = TRUE),
            SD = sd(data$AVAL, na.rm = TRUE),
            Geo_Mean = 'NE'
        )
    } else {
        list(
            Mean = 'ND',
            Median = median(data$AVAL, na.rm = TRUE),
            Maximum = max(data$AVAL, na.rm = TRUE),
            Min = 'ND',
            SD = 'ND',
            Geo_Mean = 'NE'
        )
    }
}

pre_dose_df <- ADPC[ADPC$NFRLT <= 0, ]
pre_dose_summary <- summary_stats_pre(pre_dose_df)

summary_stats_post <- function(data) {
  ltr_blq_ratio <- length(data$AVALCAT1 %in% c('BLQ','LTR','<PCLLOQ')) / nrow(data)
  
  if (ltr_blq_ratio <= 1/3) {
    list(
[...]

PR #1048, opened 2023-08-31
Add helper functions to refactor PKCT01

### LINKED ISSUE #277: PKCT01 to tern

edelarua:
Closes #277 #543 

In this PR: 

- Added `imputation_rule` function to perform 1/3 and 1/2 imputation rule calculations.
- Added `format_sigfig` formatting function to format numeric values by number of significant digits.
- Updated `analyze_vars_in_cols` to add `.aligns` argument, option to implement imputation rule (`imp_rule`), and use caching.
- Updated `add_rowcounts` to allow addition of row counts from `alt_counts_df` via the `alt_counts` argument.


R/analyze_vars_in_cols.R
```diff
@@ -206,20 +203,21 @@ analyze_vars_in_cols <- function(lyt,
     lyt <- split_cols_by_multivar(
       lyt = lyt,
       vars = vars,
-      varlabels = .labels[.stats],
-      extra_args = list(cache_env = replicate(length(.stats), list(env)))
+      varlabels = .labels[.stats]
     )
   }
 
+  env <- new.env() # create caching environment
```
Melkiades, line 210: Do we still need to cache the env? I do not remember why we added it

edelarua: I just added that in because otherwise it calculates every statistic for every cell. Using caching it will only calculate every statistic once per row.

Melkiades: Ah yes now I remember, it was needed for "computing time"

R/analyze_vars_in_cols.R
```diff
@@ -201,17 +207,36 @@ analyze_vars_in_cols <- function(lyt,
...
+
   if (do_summarize_row_groups) {
     if (length(unique(vars)) > 1) {
       stop("When using do_summarize_row_groups only one label level var should be inserted.")
```
Melkiades, line 214: I am so skeptical about using this duplication these days. We will have to change this in the future

R/analyze_vars_in_cols.R
```diff
@@ -201,17 +207,36 @@ analyze_vars_in_cols <- function(lyt,
...
-        function(u, .spl_context, labelstr, ...) {
+      function(stat, cache_env) {
+        function(u, .spl_context, labelstr, .df_row, ...) {
           # Statistic
-          res <- s_summary(u, ...)[[stat]]
+          var_row_val <- paste(
+            gsub("\\._\\[\\[[0-9]+\\]\\]_\\.", "", paste(tail(.spl_context$cur_col_split_val, 1)[[1]], collapse = "_")),
+            paste(.spl_context$value, collapse = "_"),
+            sep = "_"
+          )
```
Melkiades, line 226: I am just wondering if there is not a simpler way to do this

R/imputation_rule.R
```diff
@@ -0,0 +1,59 @@
...
+      if (!post && !stat %in% c("median", "max")) val <- NA # 1/3_pre_GT
+      if (post && !stat %in% c("median", "max", "geom_mean")) val <- NA # 1/3_post_GT
+    }
+  } else if (imp_rule == "1/2") {
+    if (ltr_blq_ratio > 1 / 2 && !stat == "max") {
+      val <- NA # 1/2_GT
+      na_level <- "ND" # 1/2_GT
+    }
+  }
+
```
Melkiades, line 57: This is a condensed version of https://github.com/insightsengineering/tern/issues/543

do you think we could test against that long list?

Melkiades (approved):
Lgtm! Great work! :) Thanks Emily

R/imputation_rule.R
edelarua: I tried to keep track from the list using the comments beside each line (e.g. `1/3_pre_LT` = 1/3 imputation rule, pre-dose, less than 1/3) but it's still a bit difficult to check each case.

R/analyze_vars_in_cols.R
edelarua: Yeah I'm not sure if there's a simpler way but the problem arises from variables that are used in multiple columns - i.e. we sometimes have just `AVAL`, sometimes will have `AVAL._[[1]]_.`, `AVAL._[[2]]_.`, `AVAL._[[3]]_.`, etc. - and then I didn't want any errors occurring if there's a `.` in a regular variable name.

R/imputation_rule.R
Melkiades: yep also for me, maybe keeping the list as a tested ref?

R/analyze_vars_in_cols.R
Melkiades: I see! It makes sense but I believe this happens only with split_by_multivars

anajens:
Hi @edelarua , please don't merge just yet. I've been using this branch to QC the table with real data (https://github.com/insightsengineering/sme-tasks/issues/184) and I am having some trouble with `.aligns`. Will tag you in the issue when I create it.

anajens (changes requested):
Here's an example of the error I see when testing this branch:

```
library(dplyr)

anl <- data.frame(
  AVAL = c(419000, 569000, 379000)
)

lyt <- basic_table() %>%
  analyze_vars_in_cols(
    vars = "AVAL",
    .stats = c("n", "mean", "sd"),
    .formats = c(n = "xx.", mean = format_sigfig(3), sd = format_sigfig(3)),
    .aligns = "decimal"
  )

build_table(lyt = lyt, df = anl)
```

Error:
```
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'x' in selecting a method for function 'toString': Decimal alignment is not supported for multiple values. Found the following string with multiple numbers (first 3 selected from column SD): '1e+05', 'NA', 'NA'
```

Can see result with `aligns = NULL`
```
       n    Mean     SD  
—————————————————————————
root   3   456000   1e+05
```

edelarua:
> Here's an example of the error I see when testing this branch:
> 
> ```
> library(dplyr)
> 
> anl <- data.frame(
>   AVAL = c(419000, 569000, 379000)
> )
> 
> lyt <- basic_table() %>%
>   analyze_vars_in_cols(
>     vars = "AVAL",
>     .stats = c("n", "mean", "sd"),
>     .formats = c(n = "xx.", mean = format_sigfig(3), sd = format_sigfig(3)),
>     .aligns = "decimal"
>   )
> 
> build_table(lyt = lyt, df = anl)
> ```
> 
> Error:
> 
> ```
> Error in h(simpleError(msg, call)) : 
>   error in evaluating the argument 'x' in selecting a method for function 'toString': Decimal alignment is not supported for multiple values. Found the following string with multiple numbers (first 3 selected from column SD): '1e+05', 'NA', 'NA'
> ```

It seems like it's caused by the combination of the formatting function (`format_sigfig`) and the decimal alignment since changing either one will fix it. 

@ayogasekaram @Melkiades do you have any insight into why this would happen with the decimal alignment? The formatting function converts to character. Strangely, this error does not occur in the PKCT01 table.

R/imputation_rule.R
```diff
@@ -0,0 +1,59 @@
...
+#'   AVALCAT1 = sample(c(1, "BLQ"), 50, replace = TRUE)
+#' )
+#' x_stats <- s_summary(df$AVAL)
+#' imputation_rule(df, x_stats, "max", "1/3")
+#' imputation_rule(df, x_stats, "geom_mean", "1/3")
+#' imputation_rule(df, x_stats, "mean", "1/2")
+#'
+#' @export
+imputation_rule <- function(df, x_stats, stat, imp_rule, post = FALSE) {
+  checkmate::assert_choice("AVALCAT1", names(df))
```
anajens, line 35: The variable name should not be hardcoded.

R/imputation_rule.R
```diff
@@ -0,0 +1,59 @@
...
+#'   AVAL = runif(50, 0, 1),
+#'   AVALCAT1 = sample(c(1, "BLQ"), 50, replace = TRUE)
+#' )
+#' x_stats <- s_summary(df$AVAL)
+#' imputation_rule(df, x_stats, "max", "1/3")
+#' imputation_rule(df, x_stats, "geom_mean", "1/3")
+#' imputation_rule(df, x_stats, "mean", "1/2")
+#'
+#' @export
+imputation_rule <- function(df, x_stats, stat, imp_rule, post = FALSE) {
```
anajens, line 34: GDS_PK_TLG_Specification (link in #543) document describes using the variable `FLAGSUM` to control the display of summary statistics. How does this fit in with this function?

edelarua: Translating from the spec. document linked in #543, in our data `SUMMCONC` is replaced with `AVAL`, `NOMTIME` is replaced by `NFRLT`, and `FLAGSUM` is `n_blq` derived within the function (sum of flagged values).

anajens: I think deriving an equivalent of `FLAGSUM` within the analysis function is risky since the code list of strings of what is considered "BLQ" is not standardized and it's possible that there are data with values other than "BLQ", "LTR", "<PCLLOQ"  (@qiliu1013 can you confirm?)

I suggest to use the variable directly from the analysis data rather than re-deriving it.

edelarua: I've updated the function to use `FLAGSUM` (not hard-coded). I'm not sure if this is a standard variable, but if not we can just revert the change.

Melkiades:
> @ayogasekaram @Melkiades do you have any insight into why this would happen with the decimal alignment? The formatting function converts to character. Strangely, this error does not occur in the PKCT01 table.

Yes, I know why exactly. It is that we should avoid using scientific notation for decimal alignments and `as.character` does it. I am preparing a fix in `formatters` to deal with this. Just a question for @anajens: is it always to avoid the scientific numeric representation?

R/formatting_functions.R
```diff
@@ -209,6 +209,33 @@ format_xx <- function(str) {
...
+#' fmt_5sf(0.57)
+#' fmt_5sf(0.000025645)
+#'
+#' @family formatting functions
+#' @export
+format_sigfig <- function(sigfig) {
+  checkmate::assert_integerish(sigfig)
+  function(x, ...) {
+    if (!is.numeric(x)) stop("`format_sigfig` cannot be used for non-numeric values. Please choose another format.")
+    as.character(signif(x, sigfig))
```
Melkiades, line 235:
```suggestion
    formatC(x, digits = sigfig, format = "f")
```

Melkiades: This should fix it

R/imputation_rule.R
Melkiades: @qiliu1013 checked the output in the final print and it seems fine! 👍

anajens:
>Just a question for @anajens: is it always to avoid the scientific numeric representation?

@Melkiades formatters should be able to handle scientific notation alignment as it's supposed to be a general purpose package and not necessarily supporting only the roche standards

Melkiades:
> > Just a question for @anajens: is it always to avoid the scientific numeric representation?
> 
> @Melkiades formatters should be able to handle scientific notation alignment as it's supposed to be a general purpose package and not necessarily supporting only the roche standards

It does handle them. Only the decimal alignment is not thinking about it (I did not when reworking it in). I was just wondering if in general they should be disabled generally. I am adding the possibility of having scientific notation and decimal alignment (i.e. transform the scientific notation in numeric, then format it

R/imputation_rule.R
edelarua: @anajens after talking to Qi I have reverted this - `FLAGSUM` is no longer a standard variable and `AVALCAT1` should be used instead.

edelarua:
@anajens regarding row counts in PKCT01 - Qi has indicated that this is a must-have so I have refactored `add_rowcounts` to suit our purposes (just need to specify `alt_counts = TRUE` if `alt_counts_df` counts should be used). See the updated snapshots which include row counts (values unchanged) [here](https://github.com/insightsengineering/tlg-catalog/pull/104/files#diff-ed139c336043416042a18ca53ef3e0b73eec9be56dc3a30b470f2bc64c62e940).

R/formatting_functions.R
edelarua: I modified this a bit but it's now working. Thanks!

R/analyze_vars_in_cols.R
```diff
@@ -201,17 +207,36 @@ analyze_vars_in_cols <- function(lyt,
...
+          if (is.null(cache_env[[var_row_val]])) cache_env[[var_row_val]] <- s_summary(u, ...)
+          x_stats <- cache_env[[var_row_val]]
+
+          if (is.null(imp_rule) || !stat %in% c("mean", "sd", "cv", "geom_mean", "geom_cv", "median", "min", "max")) {
+            res <- x_stats[[stat]]
+          } else {
+            res_imp <- imputation_rule(
+              .df_row, x_stats, stat,
+              imp_rule = imp_rule, post = as.numeric(tail(.spl_context$value, 1)) > 0
```
anajens, line 235: Please add `avalcat_var` here so user can select a different value from the default AVALCAT1 if necessary.

anajens (commented):
I'm seeing something very weird with re-using the same layout with `analyze_vars_in_cols` on different data sets. @edelarua Can you replicate this?

```
anl1 <- data.frame(AVAL = c(100, 200, 300))
anl2 <- data.frame(AVAL = c(1, 2, 3))

lyt <- basic_table() %>%
  analyze_vars_in_cols(
    vars = "AVAL",
    .stats = c("n", "mean", "sd"),
    .formats = c(n = "xx.", mean = "xx.x", sd = "xx.x"))

build_table(lyt = lyt, df = anl1)
build_table(lyt = lyt, df = anl2)

```
Tables are identical?! 
```
> build_table(lyt = lyt, df = anl1)
       n   Mean     SD  
————————————————————————
root   3   200.0   100.0
> build_table(lyt = lyt, df = anl2)
       n   Mean     SD  
————————————————————————
root   3   200.0   100.0
```

edelarua:
> I'm seeing something very weird with re-using the same layout with `analyze_vars_in_cols` on different data sets. 

Hi @anajens,

I noticed that as well - it was a side effect of caching while using the same layout. It is fixed now (by https://github.com/insightsengineering/tern/pull/1048/commits/cbfb9aa095fa5a0da13dda7f3c4aaaccaeb9be4d) since I turned off caching by default. I have also exposed the `avalcat_var` argument in `analyze_vars_in_cols`.

R/analyze_vars_in_cols.R
```diff
@@ -8,6 +8,17 @@
 #'
 #' @inheritParams argument_convention
 #' @inheritParams rtables::analyze_colvars
+#' @param imp_rule (`character`)\cr imputation rule setting. Defaults to `NULL` for no imputation rule. Can
+#'   also be `"1/3"` to implement 1/3 imputation rule or `"1/2"` to implement 1/2 imputation rule. In order
+#'   to use an imputation rule, the `AVALCAT1` variable must be present in the data. See [imputation_rule()]
+#'   for more details on imputation.
```
anajens, line 14:
```suggestion
#'   to use an imputation rule, the `avalcat_var` argument must be specified. See [imputation_rule()]
#'   for more details on imputation.
```

anajens (approved):
Thanks for all the updates @edelarua

R/formatting_functions.R
Melkiades: @edelarua just for better understanding, why did you modify it specifically? I am reading the docs but I find hard figuring out use cases ^^

PR #1067, opened 2023-09-21
Fix trailing "." when using `format_sigfig`, add functionality for multiple values

### LINKED ISSUE #1066: Fix bug in `format_sigfig` causing trailing "."
For some numbers, a trailing "." is included after formatting that should be removed, eg. 200.0 to "200."

edelarua:
Closes #1066


NEWS.md
```diff
@@ -14,6 +14,9 @@
 * Updated `add_rowcounts` to allow addition of row counts from `alt_counts_df` using the `alt_counts` argument.
 * Added `gp` argument to `g_forest` to control graphical parameters such as font size.
 
+### Bug Fixes
+* Fixed bug in `format_sigfig` causing some numbers to be printed with trailing "." character.
```
anajens, line 18: format_sigfig is new function so no need to capture this

anajens (approved):
Thanks @edelarua !

R/formatting_functions.R
```diff
@@ -228,11 +231,21 @@ format_xx <- function(str) {
...
 #' @export
-format_sigfig <- function(sigfig) {
+format_sigfig <- function(sigfig, type = c("value", "value_paren", "range")) {
   checkmate::assert_integerish(sigfig)
+  type <- match.arg(type)
   function(x, ...) {
     if (!is.numeric(x)) stop("`format_sigfig` cannot be used for non-numeric values. Please choose another format.")
-    formatC(signif(x, digits = sigfig), digits = sigfig, format = "fg", flag = "#")
+    num <- formatC(signif(x, digits = sigfig), digits = sigfig, format = "fg", flag = "#")
+    num <- gsub("\\.$", "", num) # remove trailing "."
```
Melkiades, line 240: I think this is not the best way to do this. Also, here there should be the possibility to have it fixed. I think flag adds the dot and g grounds. There is no need of signif function here from my testing. I would keep this as another option and remove the value value_paren and range as this is not yet needed and this will be a core function for the future. Composite functions can be added in a different way (by parsing and replacing) which is done in one of the formatting functions I have added. We can talk about this at the SU

Melkiades (changes requested):

edelarua:
Notes on implementation:
- Format `"fg"` is used to format using significant digits - `"f"` prints with fixed (non-scientific notation) format and `"g"` causes `digits` to represent number of significant digits (not decimal places).
- Without the `"#"` flag, additional required digits are not correctly added and leading whitespace is added instead. For example:
```
> formatC(signif(1.5, digits = 3), digits = 3, format = "fg", flag = "#")
[1] "1.50"
> formatC(signif(1.5, digits = 3), digits = 3, format = "fg")
[1] " 1.5"
```
- Without `sigfig` values are not rounded correctly. For example:
```
> formatC(signif(123123, digits = 3), digits = 3, format = "fg")
[1] "123000"
> formatC(123123, digits = 3, format = "fg")
[1] "123123"
```

anajens (changes requested):
@edelarua this is looking a lot better but I think we need to be careful about which formats are supported. I suggest to either restrict the format list more explicitly and/or add more tests to verity that all formats in `list_valid_format_labels()` are supported:

Here's 2 cases to think about, both seem to be bugs in `format_value`:

(1) this gives an error:
```
num1 <- c(0.1239)
z <- format_sigfig(3, format = "xx%" )
z(num1)
>Error in x * 100 : non-numeric argument to binary operator
```

(2) This gives the wrong result "<0.0001"
```
num2 <- c(1234)
z <- format_sigfig(3, format = "x.xxxx | (<0.0001)" )
z(num2)
```

anajens (approved):
Thanks, this is much safer 😅

PR #1075, opened 2023-10-02
Add formatting function with special case when count less than 10

### LINKED ISSUE #1069: Suppress display of % if number of observations is less than 10 for example 5 (3.2%) should display just 5.

edelarua:
# Pull Request

Fixes #1069


edelarua:
Does this format also need to use a fixed single decimal place?

Melkiades (commented):
Nice job! I feel they will ask this for "auto", but for now we can add the digits in some way? what do you think?

edelarua:
@Melkiades I feel like since it's for `count_fraction` they may not want to use it with `auto`, since it wouldn't make sense to use decimal places for a count value (not sure about the percentage DPs). Is there someone we can ask to confirm the correct format/usage?

Melkiades (approved):
Lgtm! if we have further specifications, we can easily change it accordingly. Thanks Emily! :)

barnett11:
This is a non-standard request we're providing here, so I cannot say for standards. From the request, I gather the behaviour should match the standards, just not having a percentage display.

ayogasekaram (approved):
lgtm!

PR #1071, opened 2023-09-28
Expose `na_str` argument in analyze/summarize functions

edelarua:
Closes #1070

PR #1079, opened 2023-10-05
Fix `s_summary` return for empty logical vectors

### LINKED ISSUE #649: [Feature Request]: a_summary.logical formatting not dealing as a_summary.factor when denom = 0
### Feature description

Hi @shajoezhu , 

When using the `summarize_vars` function, I realized that when the denominator is 0, `summarize_vars()` returns `NA` for logical variables while it returns `0` for factor variables for `.stats = count_fraction`. I feel both should act in the same way. I am proposing the solution. 

See example: 

```
df <- data.frame(var1 = c(TRUE, TRUE, FALSE, FALSE),
                 var2 = c("d", "d", "e", "e"),
                 var3 = c("f", "f", "f", "f"))

df <- df %>% dplyr::mutate_if(is.character,as.factor) %>%
  mutate(var3 = factor(var3, levels = c("f", "g")))

my_table <- basic_table() %>%
  split_cols_by("var3") %>%
  summarize_vars(
    vars = c("var1", "var2")
    ) %>%
  build_table(df = df)
```
Look at the red and green circles. 

image

The following is the reason why this is happening: 

- In `s_summary.logical`
`y$count_fraction <- c(count, ifelse(dn > 0, count / dn, NA))`
- In `s_summary.factor`
 ```
y$count_fraction <- lapply(
    y$count,
    function(x) {
      c(x, ifelse(dn > 0, x / dn, 0))
    }
  )
```
In the logical part we are assigning the NA value if denom is 0, In the factor case we are assigning the 0 value for the same case. Then, if we go to [format_count_fraction()](https://github.com/insightsengineering/tern/blob/main/R/formats.R) which is the function used for formatting, if `x[1]` or `x[2]` is `NA` (which is the case for the logical scenario above), then the output will be `NA`. 

So the solution would be: 

- To act in the same in both cases. I would change s_summary.logical to be aligned with s_summary.factor: 
`y$count_fraction <- c(count, ifelse(dn > 0, count / dn, NA))`

- **Don't you think that when showing counts and percentages (xx (xx%)), would be even better the output to be i.e 0/0%, instead of just 0 when denom is 0?**

See the new proposal on format_count_fraction() code: 

```
...
result <- if (x[1] == 0 && x[2] != 0) {
  "0"
} else if (x[2] == 0) {
  "0/0%"
} else {
    paste0(x[1], " (", round(x[2] * 100, 1), "%)")
  }...
```

### Code of Conduct

### Contribution Guidelines

### Security Policy

edelarua:
# Pull Request

Fixes #649

Outputting `NE` instead of `NA` will be enabled by https://github.com/insightsengineering/tern/pull/1071.


Melkiades (approved):
Lgtm! Thanks, Emily for rectifying this!

PR #1087, opened 2023-10-16
Add ref_group_last for custom split funs

### LINKED ISSUE #1085: [Feature Request]: Add split function to enable placing reference arm last in efficacy layout
### Feature description

Similar to https://github.com/insightsengineering/rtables/issues/707

### Code of Conduct

### Contribution Guidelines

### Security Policy

anajens:
Fixes #1085

Please note this is a partial solution and `ref_group_last` will not work when there are combination facets in the layout.

Once https://github.com/insightsengineering/rtables/issues/747 is closed, `ref_group_last` needs to be updated.


anajens:
@BFalquet here is the custom split function that would be good to add to chevron layouts RSPT01 and TTET01

anajens:
Blocked until team decides on whether fix should be done only at the rtables level: https://github.com/insightsengineering/rtables/pull/748#issuecomment-1764692535

R/utils_split_funs.R
```diff
@@ -0,0 +1,154 @@
...
+#'     vars = "AVAL",
+#'     var_labels = "Survival Time (Months)",
+#'     is_event = "is_event",
+#'   ) %>%
+#'   build_table(df = adtte_f)
+#'
+#' @export
+ref_group_last <- make_split_fun(
+  post = list(
+    function(splret, spl, fulldf) {
```
Melkiades, line 59: @anajens I think understanding these three parameters is a bit too much for someone who wants only to put ref_group as last so I simplified it this way. Let me know what you think!

R/utils_split_funs.R
```diff
@@ -0,0 +1,154 @@
...
+#'
+#' Collection of useful functions that are expanding on the core list of functions
+#' provided by `rtables`. See `?rtables::custom_split_funs` and [rtables::make_split_fun()]
+#' for more information on how to make a custom split function. All these functions
+#' work with [split_rows_by()] argument `split_fun` to modify the way the split
+#' happens.
+#'
+#' @seealso [rtables::make_split_fun()]
+#'
+#' @name utils_split_funs
```
Melkiades, line 13: @anajens changed the name of the file so we can add more custom split functions. Feel free to add others or change the name in anything more fitting (maybe custom_split_funs?)

R/utils_split_funs.R
```diff
@@ -0,0 +1,156 @@
...
+#'
+#' @export
+ref_group_last <- make_split_fun(
+  post = list(
+    function(splret, spl, fulldf) {
+      if (!"ref_group_value" %in% slotNames(spl)) {
+        stop("Reference group is undefined.")
+      }
+
+      spl_var <- spl@payload # can be substituted by splret$labels
```
Melkiades, line 64: @anajens @edelarua I think these (also after spl@ref_group_value) can be substituted by splret getters or $. In general, Gabe strongly discouraged using direct accessors to objects but to add and use getter more often. For me now, this is fine, as it is only an utility function

tests/testthat/test-utils_split_fun.R
```diff
@@ -0,0 +1,168 @@
...
+      method = "both"
+    ) %>%
+    build_table(adtte_f)
+
+  res <- testthat::expect_silent(result)
+  testthat::expect_snapshot(res)
+})
+
+testthat::test_that("summarize_ancova works as expected with ref_group_last split fun", {
+  anl <- tern_ex_adtte %>% filter(PARAMCD == "TNE")
```
Melkiades, line 133: I do not think we need to test so many tern functions @anajens, also because the split_function is working on split_*_by and not in tern functions

tests/testthat/test-utils_split_fun.R
```diff
@@ -0,0 +1,168 @@
...
+})
+
+# testthat::test_that("ref_group_last split fun gives error when used with combo facets", {
+#   custom_splitfun <- make_split_fun(
+#     post = list(
+#       add_combo_facet("A_C", "Arms A+C", c("A: Drug X", "C: Combination")),
+#       ref_group_last
+#     )
+#   )
+#
```
Melkiades, line 16: @anajens I think this is very legitimate and if you think there is added value in having ref_group_last as post split_fun we can put it back imo. I would add anyway the current shorthand as I do not expect many people to use composite split_fun s

R/utils_split_funs.R
Melkiades: also there is this small issue: 
```
Undocumented arguments in documentation object 'utils_split_funs'
  ‘df’ ‘spl’ ‘vals’ ‘labels’ ‘trim’ ‘.spl_context’
```
I guess these can be imported from {rtables} or argument_conventions. Do you think this is reasonable @edelarua?

edelarua: Yeah I would do `@inheritParams` for those

R/utils_split_funs.R
anajens: This looks good to me

R/utils_split_funs.R
anajens: @Melkiades yes this is OK and probably more convenient for users. Initially I was thinking to export just the core function because that way users could reuse it in the their own split functions.

R/utils_split_funs.R
anajens: Absolutely! I tried to use the getter but it was exported from rtables so thats why I used @

tests/testthat/test-utils_split_fun.R
anajens: This was more of a sanity check for me to see if there are any edge cases is layouts that I didn't consider. I'm OK to shorten the list but I think we should test it on at least 1 tern layout function.

R/utils_split_funs.R
```diff
@@ -0,0 +1,156 @@
...
+      init_lvls <- levels(fulldf[[spl_var]])
```
anajens, line 65: TODO: add here code for the case when the split var is not a factor.

Melkiades (approved):

edelarua (approved):
Lgtm!

PR #1089, opened 2023-10-16
1086 update plot snaps@main

### LINKED ISSUE #1086: Update tests for plots to image snaps.
like [this example](https://github.com/insightsengineering/tern/tree/main/tests/testthat/_snaps/g_km)

- [x] g_km
- [x] g_lineplot
- [x] g_step 
- [x] g_forest
- [x] g_waterfall

ayogasekaram:
closes #1086


edelarua:
@ayogasekaram did you run the tests for `g_km` via the "Test" button/`devtools::test()`? Just wanted to check, because if they're run via `devtools::test_active_file` (or similar) they produce very minor changes in the snapshots and will fail integration tests. This is because some of the `g_km` tests use annotations that are sized based on viewport size which is slightly different between these different test functions. Should be fine for all other tests as far as I know!

ayogasekaram:
@edelarua Thanks for letting me know. I just reran the tests with devtools::test and the g_km plots changed but they should reflect the correct size now.

PR #1120, opened 2023-11-02
Add option to set reference arm in `g_km`

### LINKED ISSUE #1119: Add option to set reference arm in `g_km`
In the coxph annotation table for `g_km`, reference arm is automatically set to the first level of the arm variable. A parameter should be added in (or added to the `control_coxph_pw` input list) to set the reference arm. We should also increase visibility in the documentation for the reference arm and/or print it in the coxph annotation table somewhere.

edelarua:
# Pull Request

Fixes #1119

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,8 @@
 * Added `level_order` split function to reorder manually the levels.
 
 ### Enhancements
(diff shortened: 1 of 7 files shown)


Melkiades (approved):
Lgtm! Thanks Emily :) just one question: is "vs." the standard?

Issue #1117, opened 2023-11-02
[Feature Request]: Improve documentation for analyze_vars()


kieranjmartin:
### Feature description

As of version 0.9.2, the help page for `analyze_vars` begins by describing in some detail `s_summary`, and gives a great deal of usage for it. If I understand correctly, I think `analyze_vars` is calling `s_summary` to do summary stats, although I'm not quite sure. Either way, I don't think this is a good use case for a shared help page, and for a core tern function, I think this makes it much more confusing as to how to properly use `analyze_vars`. In particular, it isn't immediately obvious which stats are usable for which object types; you can sort of work it out by looking at the returned list for each class, but that's a very roundabout way of delivering the information!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

edelarua:
Sounds good! I'll make a separate PR to tidy up the documentation.

PR #1123, opened 2023-11-02
Expand `get_stats` to cover all analysis functions

### LINKED ISSUE #1108: Expand `get_stats` to cover all analysis functions
Currently `get_stats` explicitly covers the following methods: `"count_occurrences"`, `"summarize_num_patients"`, `"analyze_vars_counts"`, and `"analyze_vars_numeric"`. This function should be expanded to cover all analysis functions (and their statistics) available in `tern`.

edelarua:
# Pull Request

Fixes #1108


R/utils_default_stats_formats_labels.R
```diff
@@ -306,7 +296,51 @@ get_indents_from_stats <- function(stats, indents_in = NULL, row_nms = NULL) {
...
 
-#' @describeIn default_stats_formats_labels Named list of default formats for `tern`.
+#' @describeIn default_stats_formats_labels Named list of available statistics by method group for `tern`.
+#'
+#' @format
+#' * `tern_default_stats` is a named list of available statistics, with each element
+#'   named for their corresponding statistical method group.
+#'
+#' @export
+tern_default_stats <- list(
```
Melkiades, line 306: This is great!! I was thinking that we could also add this info throughout the documentation and replace the defaults with this in the formals. What do you think?

Melkiades: I think this would solve also https://github.com/insightsengineering/tern/issues/1117

edelarua: @Melkiades I'll definitely add this info throughout the documentatation ("available stats for this function are..."), but for most functions we wouldn't want to replace the `.stats = ...` statistics, correct? Since `get_stats()` returns all possible statistics for the method/function and `.stats` should only list the default(s) used. Correct me if this is wrong. E.g. `get_stats("analyze_vars_numeric")` lists a lot more stats than we want when we use `analyze_vars` with the default settings.

I think we should also discuss moving sfuns to their own documentation pages again since we'll have to do this anyways when `tern.s` is ready and this would help with #1117.

Melkiades: Yes it makes sense. I will answer the issue according to this, thank you :)

R/response_biomarkers_subgroups.R
```diff
@@ -72,7 +72,7 @@ tabulate_rsp_biomarkers <- function(df,
   checkmate::assert_data_frame(df)
   checkmate::assert_character(df$biomarker)
   checkmate::assert_character(df$biomarker_label)
-  checkmate::assert_subset(vars, c("n_tot", "n_rsp", "prop", "or", "ci", "pval"))
+  checkmate::assert_subset(vars, get_stats("tabulate_rsp_biomarkers"))
```
Melkiades, line 75: This is great in my opinion

Melkiades (approved):
Lgtm! Thanks, Emily for this!! :) I think now it is much clearer what stats can be used with what

PR #1129, opened 2023-11-07
Improve `analyze_vars` documentation

### LINKED ISSUE #1117: [Feature Request]: Improve documentation for analyze_vars()

edelarua:
# Pull Request

Fixes #1117

We may want to consider moving all analyze functions to the tops of their resepective documentation pages for better visibility. I could also improve the function description for other documentation pages if we want to improve documentation throughout the whole package.


Melkiades (approved):
This is amazing!! Much more readable and linear, easy to understand. I advocate doing this for every summary function, but I know it is a lot of work, so we can do it at a later time if you prefer to work on something else :)

Issue #1130, opened 2023-11-08
Improve readability of analyze function documentation


edelarua:
Similar to https://github.com/insightsengineering/tern/pull/1129, make it easier for the user to read & find information about the analyze function on shared documentation pages. Since statistics/analysis functions are not intended for use by the average `tern` user, we can give the analyze functions precendence in documentation pages by assigning them the `@order 1` roxygen tag.

Melkiades:
Thanks for this Emily!! Reading your comment here, I was thinking that we might add the list of summary functions to the README and intro vignette so to show users how to check if their functionality is already somehow present. What do you think?

edelarua:
Sounds good! I think any vignette changes should definitely be in a separate issue if you want to create one for that.

Melkiades:
@edelarua fantastic job!! :)

PR #1131, opened 2023-11-10
Reorder doc pages so analyze functions appear first, add links in README

edelarua:
# Pull Request

Part of #1130

I'm separating this issue into multiple PRs so nothing gets mixed up.


R/abnormal.R
```diff
@@ -35,20 +36,7 @@ NULL
...
-#'   USUBJID = as.character(c(1, 1, 2, 2)),
-#'   ANRIND = factor(c("NORMAL", "LOW", "HIGH", "HIGH")),
-#'   BNRIND = factor(c("NORMAL", "NORMAL", "HIGH", "HIGH")),
-#'   ONTRTFL = c("", "Y", "", "Y"),
-#'   stringsAsFactors = FALSE
-#' )
 #'
-#' # Select only post-baseline records.
-#' df <- df %>%
-#'   filter(ONTRTFL == "Y")
```
Melkiades, line 51: Just wondering if this pre-proc should be in the intro of the doc page, i.e. where NULL is. So it is a bit more readable codewise

R/abnormal_by_baseline.R
```diff
@@ -29,6 +29,7 @@
 #' @seealso Relevant description function [d_count_abnormal_by_baseline()].
 #'
 #' @name abnormal_by_baseline
+#' @order 1
```
Melkiades, line 32: do you need to have order 1 for the main NULL part?

README.md
```diff
@@ -21,27 +21,35 @@
...
-    -   ...
--   Summary tables:
-    -   unique patients
-    -   exposure across patients
-    -   change from baseline for parameters
-    -   ...
+
+Data visualizations:
+
+-   Line plots ([`g_lineplot`](https://insightsengineering.github.io/tern/main/reference/g_lineplot.html))
```
pawelru, line 30: Hey @m7pr, IIRC you were involved in the discussion of changing `/main/` into `/latest-tag/`. Did it happened? If yes, can you please provide a rationale for folks and assure it follows standards?

m7pr: It basically was that `/main/` can have changes and features that were not released, and our front landing website supposed to have the most recent released version of our software and the documentation, hence that's why the latest-tag (released version).

m7pr: + https://insightsengineering.github.io/package_name/ redirects to https://insightsengineering.github.io/package_name/latest-tag/ so this is standard across all GitHub, that they display the documentation of most recent release

R/abnormal_by_baseline.R
edelarua: Yes, otherwise parameters defined here can be overridden, some other pieces of documentation are out of order, etc.

R/abnormal.R
edelarua: Sure, I can move it. I figured for functions that haven't been refactored yet/don't have examples added except for their analyze functions it might be easier for us to run the code just from one place. But it makes sense to move the data generation step to the other part for functions with examples in multiple places. What do you think?

README.md
edelarua: Thanks @pawelru @m7pr - I will remove the `/main/` part from all URLs here!

R/abnormal.R
Melkiades: I see. We can enforce a better standard in the future, but thinking about it now, it makes more sense to keep it as you modified it, also considering that we will remove stats one day.

README.md
m7pr: amazing

README.md
```diff
@@ -21,27 +21,35 @@
...
+Statistical model fit summaries:
+
+-   Logistic regression ([`summarize_logistic`](https://insightsengineering.github.io/tern/reference/summarize_logistic.html))
+-   Cox regression ([`summarize_coxreg`](https://insightsengineering.github.io/tern/reference/cox_regression.html))
+
+Analysis tables:
+
+-   See a list of all available analyze functions [here](https://insightsengineering.github.io/tern/reference/analyze_functions.html)
+-   See a list of all available summarize functions [here](https://insightsengineering.github.io/tern/reference/summarize_functions.html)
+-   See a list of all available column-wise analysis functions [here](https://insightsengineering.github.io/tern/reference/analyze_colvars_functions.html)
```
Melkiades, line 46: Thanks for this. It was really needed!

R/survival_biomarkers_subgroups.R
```diff
@@ -29,62 +29,11 @@
...
-#'     strata = "STRATA1",
-#'     covariates = "SEX",
-#'     subgroups = "BMRKR2"
-#'   ),
-#'   data = adtte_f
-#' )
-#' df
-#'
-#' @name survival_biomarkers_subgroups
-NULL
```
Melkiades, line 46: the examples were repeated right?

R/survival_duration_subgroups.R
```diff
@@ -78,61 +96,6 @@ NULL
...
-#' @examples
-#' library(dplyr)
-#' library(forcats)
-#'
-#' adtte <- tern_ex_adtte
-#' adtte_labels <- formatters::var_labels(adtte)
-#'
-#' adtte_f <- adtte %>%
-#'   filter(
-#'     PARAMCD == "OS",
```
Melkiades, line 90: this pre-proc was too repeated right?

Melkiades (approved):
Amazing job Emily!! Thanks so much for polishing and consolidating examples too. I think it will feel much fresher now. Also having the main {rtables}/user-facing function top is much better now. Do you think we should also add a seealso for the lists of analysis functions? 

Good to go anyway ;)

edelarua:
> Do you think we should also add a seealso for the lists of analysis functions?

I don't think it's necessary to link to all the other analyze functions since most of them are not closely related - any that are can be linked via `seealso`, and if someone isn't sure what they're looking for they would hopefully visit the analyze/summarize functions page directly from the pkgdown site home/reference page.

PR #1134, opened 2023-11-16
Expose statistic function arguments in analyze functions

### LINKED ISSUE #1133: Expose statistic function arguments in analyze functions
Most analyze functions currently require arguments that are not explicitly listed in their arguments. These arguments (as well as optional settings specific to the function) should be exposed to the user so that they know what is required to use the function. This way users are not also required to read the documentation for the associated statistic function when using an analyze function and error messages relating to these arguments are more coherent.

For example, the `thresholds` argument is required to use `count_cumulative` but this argument is not listed in the usage section for `count_cumulative`, only for `s_count_cumulative`.

This update will also be useful later on when refactoring analyze functions and when statistics functions are moved to the tern.s package.

Part of https://github.com/insightsengineering/tern/issues/1130

edelarua:
# Pull Request

Fixes #1133

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-# tern 0.9.2.9015
+# tern 0.9.2.9016
 
 ### New Features
 * Added `ref_group_position` function to place the reference group facet last, first or at a certain position.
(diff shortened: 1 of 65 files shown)


ayogasekaram (approved):
lgtm!

Issue #850, opened 2023-03-10
Add a Bland-Altman analysis function


danielinteractive:
To do:

- [x] Add a function `s_bland_altman(x, y, conf_level = 0.95, group = NULL)`
- [x] Make the result a `bland_altman` S3 class list with elements
	- [ ] `difference_mean`
	- [ ] `ci_mean` 
	- [ ] `difference_sd`
	- [ ] `difference_se`
	- [ ] `upper_agreement_limit` 
    - [ ] `lower_agreement_limit` 
    - [ ] `agreement_limit_se`
    - [ ] `upper_agreement_limit_ci`
    - [ ] `lower_agreement_limit_ci`
    - [ ] `t_value`
    - [ ] `n`
- [x] Add tests
- [ ] Add to NEWS
- [ ] Add to `pkgdown` yaml file

PR #1135, opened 2023-11-20
Adding Bland- Altman statistical function

linc50:
# Pull Request

I have added the Bland-Altman sfunction, which can be used to assess the agreement between two quantitative methods of measurement. Please kindly help to review the code.

Thanks,
Chanjuan


linc50:
close issue #850

tests/testthat/test-bland-altman.R
```diff
@@ -0,0 +1,56 @@
...
+  )
+  expect_identical(res, expect, tolerance = 1e-5)
+
+
+})
+
+
+testthat::test_that("works correctly with NA element in either vectors", {
+  set.seed(1)
+  x = rnorm(20)
```
Melkiades, line 37: Use `<-`

tests/testthat/test-bland-altman.R
```diff
@@ -0,0 +1,56 @@
...
+    lower_agreement_limit = -2.093733,
+    agreement_limit_se = 0.5393764,
+    upper_agreement_limit_ci = c(1.555070, 3.420377),
+    lower_agreement_limit_ci = c(-3.026386,-1.161079),
+    t_value = 1.729133,
+    n = 20L
+  )
+  expect_identical(res, expect, tolerance = 1e-5)
+
+
```
Melkiades, line 31: not needed

R/bland_altman.R
```diff
@@ -0,0 +1,73 @@
+
```
Melkiades, line 1: not needed

R/bland_altman.R
```diff
@@ -0,0 +1,73 @@
...
+#' Bland Altman analysis
+#'
+#' @description 'r lifecycle::badge("stable")'
+#'
+#' @inheritParams argument_convention
+#' @param y ('numeric')\cr vector of numbers we want to analyze.
```
Melkiades, line 7: are you sure x is defined correctly in argument convention?

R/bland_altman.R
```diff
@@ -0,0 +1,73 @@
+
+#' Bland Altman analysis
+#'
+#' @description 'r lifecycle::badge("stable")'
```
Melkiades, line 4: Description is not sufficient and it is missing connection with its main `rtables` uses

R/bland_altman.R
```diff
@@ -0,0 +1,73 @@
...
+#'
+#' @inheritParams argument_convention
+#' @param y ('numeric')\cr vector of numbers we want to analyze.
+#'
+#' @name s_bland_altman
```
Melkiades, line 9: I am not sure we need a dedicated file for this function, as we have already so many

Melkiades (changes requested):
Thanks @linc50 for working on this :)

We need a NEWS entry, the documentation is not present (you need to roxygenize and add description and fix the reference), and we need to have this statistical function is some tabulation example, test and list of stats (see `get_stats`). I would also consider moving this to a file that has similar statistical functions as the name itself does necessarily cut it for a self-explanatory file. Remember to use the styler btw ;)

R/bland_altman.R
```diff
@@ -0,0 +1,73 @@
...
+#' @details
+#' https://doi.org/10.1016/S0140-6736(86)90837-8
+#' @references
+#' @article{bland1986statistical,
+# title={Statistical methods for assessing agreement between two methods of clinical measurement},
```
Melkiades, line 14: bibliography is handled in `tern` so to have bib refs in a dedicated file avoiding the style/linter exceptions and it is clearer to read. Search for the .bib file and its content to see how we accomplish this

R/bland_altman.R
linc50: yes, x is the vector we want to analyze, and to make it clear I added more description of parameter y.

R/bland_altman.R
linc50: Given there might be other functions applying bland altman method (for example,  bland altman plot), we'd better to keep this dedicated file for further development.

R/bland_altman.R
linc50: have moved the bib information to REFERENCES.bib under tern

R/bland_altman.R
linc50: added more information for description. And to be honest, this function does have little connection with rtables, shall we change the location of this function?

R/bland_altman.R
```diff
@@ -39,14 +39,20 @@ s_bland_altman <- function(x, y, conf_level = 0.95){
   ind <- complete.cases(x, y) # use only pairwise complete observations, and check if x and y have the same length
   x <- x[ind]
   y <- y[ind]
+  n <- length(n)                                      # number of 'observations'
```
Melkiades, line 28: `length(ind)` maybe?

R/bland_altman.R
```diff
@@ -39,14 +39,20 @@ s_bland_altman <- function(x, y, conf_level = 0.95){
...
+
+  if(n ==0){
```
Melkiades, line 30:
```suggestion
  if(n == 0){
```

R/bland_altman.R
Melkiades: We cannot have a dedicated file for every method only for further development. If its documentation and file remain stand-alone, no other will ever see or use it. I suggest adding it to some group of statistics/methods and adding links to it in the current documentation (remember to add it to pkgdown too)

shajoezhu:
Hi @Melkiades , we will update this PR with a g_bland_altman function

R/bland_altman.R
Melkiades: How do you use this function? Do you generate tables/plots with it?

R/bland_altman.R
Melkiades: I see it works, but maybe a direct ref when mentioned would be beneficial

Melkiades:
> Hi @Melkiades , we will update this PR with a g_bland_altman function

Ok super! so far, I think we have used the main user-facing (tables/plot) functions to name and direct files' content. I would probably wait to have the plot function and name it as you suggested, `g_bland_altman`. We can always refactor stats and methods to have a different semantic grouping at a later time (also after Daniels' work maybe?)

R/bland_altman.R
pawelru: please have a look at the pre-commit tool that will does that for you automatically on commit

tests/testthat/test-bland-altman.R
```diff
@@ -0,0 +1,54 @@
...
+  testthat::expect_error(s_bland_altman(x = 1:5, y = 1:6, 0.95))
+})
+
+testthat::test_that("infeasible input gives correct error", {
+  testthat::expect_error(s_bland_altman(x = c("a", "b", "c"), y = 1:3, 0.95))
+  testthat::expect_error(s_bland_altman(x = 1:3, y = 4:6, 2))
+})
+
+
+testthat::test_that("works correctly", {
```
pawelru, line 11: Can you please enhance test description. This doesn't tell you much :) 
Note that these descriptions are visible on a package level when running tests for the whole package.

shajoezhu (approved):
lgtm! hey @Melkiades , I was wondering if you would want to take another look, Thanks

Melkiades (commented):
@shajoezhu, we can proceed with this, but I would appreciate a little clarification. There is no primary `a_bland_altman` function or some rtables-related function. I understand this could be part of a plan to add statistical functions for other packages or for other uses. If this is the case, I suggest we group them in a file (e.g. `general_stats` or `external_stats`) where we specify that they are for external uses; I expect more functions will gather there.

On the technical side, it is missing the function references in pkgdown and needs the plot function test.

Melkiades:
> > @shajoezhu, we can proceed with this, but I would appreciate a little clarification. There is no primary `a_bland_altman` function or some rtables-related function. I understand this could be part of a plan to add statistical functions for other packages or for other uses. If this is the case, I suggest we group them in a file (e.g. `general_stats` or `external_stats`) where we specify that they are for external uses; I expect more functions will gather there.
> > On the technical side, it is missing the function references in pkgdown and needs the plot function test.
> 
> Thanks @Melkiades , I created an issue #1175 to follow up on the ggplot test, let's do this after we resolve the svg plot testing issue.
Thanks 👍 
> 
> the s_ function was dedicated for supporting the plot, there is no need for table at this moment, we had looked up on the standard before, and there was no example there.
I see, then LGTM! 👍 
> 
> I thought g_ pattern should allow us to add this to pkgdown site, correct me if I was wrong. probably no need to duplicate materials on the s_bland_altman function. thanks
Good point. It is true ^^, hence the absence of error in ci/cd

Melkiades (approved):
LGTM! Thanks for working on this ;)

PR #1182, opened 2024-01-24
options for strict tests; few enhancements

pawelru:


tests/testthat/_snaps/analyze_variables.new.md
```diff
@@ -0,0 +1,1465 @@
+# s_summary return NA for x length 0L
```
Melkiades, line 1: I think it is better to avoid pushing .new.md files, otherwise we overload the git tree ;)

tests/testthat/test-compare_variables.R
```diff
@@ -53,7 +53,7 @@ testthat::test_that("s_compare for factor handles explicit NAs as expected", {
     na.rm = FALSE
   ))
 
-  res <- testthat::expect_silent(result_with_na$pval)
+  res <- testthat::expect_silent(result_with_na$pval_counts)
```
Melkiades, line 56: Very good point. I think @edelarua we forgot to go through these when I changed the default names of pval outputs

tests/testthat/test-logistic_regression.R
```diff
@@ -109,15 +113,19 @@ testthat::test_that("fit_logistic works with two stratification variables", {
...
-    variables = list(
-      response = "Response",
-      arm = "ARMCD",
-      covariates = c("RACE", "AGE"),
-      strata = c("STRATA1", "STRATA2")
-    )
-  ))
+  # https://github.com/therneau/survival/issues/240
+  withr::with_options(
+    opts_partial_match_old,
```
Melkiades, line 118: eheh good catch!!

tests/testthat/_snaps/analyze_variables.new.md
pawelru: oh no - my mistake!

pawelru:
hey @insightsengineering/nest-sme 
When working at this I have encounter one thing that I would like to highlight to you and ask for a confirmation that that's correct

https://github.com/insightsengineering/tern/blob/01e55e6fd2c463374c6fc36ab229e666baa08f17/R/h_response_subgroups.R#L336-L340

There are also few more places with `variables$strat`:
image

which is not so consistent with `variables$strata` found here:
image

(note that my search query is imperfect)

Is this intended?

This had also downstream consequences in chevron and possibly other pkgs.

edelarua:
> > Is this intended?
> > This had also downstream consequences in chevron and possibly other pkgs.
> 
> It looks like a typo to me, but I think @edelarua and @ayogasekaram know more about strata in {tern}

Yes this is likely a typo, but comes from older code so I can't be sure if it was unintentional - it would be nice to fix this and standardize to `strata`.

R/h_response_biomarkers_subgroups.R
```diff
@@ -53,7 +53,7 @@ h_rsp_to_logistic_variables <- function(variables, biomarker) {
     response = variables$rsp,
     arm = biomarker,
     covariates = variables$covariates,
-    strata = variables$strat
+    strata = variables$strata %||% variables$strat # temporary until https://github.com/insightsengineering/tern/issues/1186 # nolint
```
pawelru, line 56: I needed to do this temporary fix here to make it green. It would be good to address the aforementioned issue before the release so that we won't be having this in the released version

edelarua: I will get to work on #1186, replacing `strat` with `strata` and adding deprecation warnings where necessary. Thanks for raising the issue!

edelarua: This is resolved now.

Melkiades (approved):
Lgtm!!! Thanks Pawel for polishing {tern} and getting the parameter correct!! :)

PR #1183, opened 2024-01-29
fix roxygenise deps; styler

pawelru:
I had issues with running pre-commit due to bioc packages but those are not needed at all!


Melkiades (approved):
Lgtm! Thanks Pawel!

PR #1192, opened 2024-02-20
fix format precision for `format_count_fraction_fixed_dp`

### LINKED ISSUE #1191: [Bug]: format_count_fraction_fixed_dp
### What happened?

A bug happened!
format_count_fraction_fixed_dp(x), when x = c(denom,denom*(1/denom)) is not always denom (100%).
This is as the check on the fraction : x[2] == 1 is not always true.

Replacing the check with signif(x[2],10^25) == 1, would resolve the bug.
```r
library(tern)
library(dplyr)

sample <- 1000

xx <- data.frame(Ncol=1:sample)
xx <- xx %>% 
  rowwise() %>% 
  mutate(count=Ncol) %>% 
  mutate(pct = count*(1/Ncol)) %>% 
  mutate(fmt_tern = format_count_fraction_fixed_dp(c(count,pct))) %>% 
  mutate(check = pct == 1) %>% 
  mutate(check2 =signif(pct,10^25) == 1) %>% 
  mutate(fmtissue = ifelse(!check,"fmt not as expected","fmt as expected")) 
  
table(xx$fmtissue)

xx[xx$count %in% c(48,49,50),]

```

### sessionInfo()

```R
tern_0.9.3.9006
```

### Relevant log output

_No response_

### Code of Conduct

### Contribution Guidelines

### Security Policy

Melkiades:
Fixes #1191


ayogasekaram (approved):
lgtm! Thanks @Melkiades

shajoezhu (dismissed):
hey @Melkiades , can you update the pr, cicd pipeline is failing, cheers

Melkiades:
I know it is silly to wonder, but I am wondering if it makes sense and if it is not too complicated to do restyling and roxygen checks one after the other so to have 2 automatic commits instead of having to do the second step manually. @cicdguy it is more difficult to explain than what you see in the commits (silly me I know)

Melkiades:
> > I know it is silly to wonder, but I am wondering if it makes sense and if it is not too complicated to do restyling and roxygen checks one after the other so to have 2 automatic commits instead of having to do the second step manually. @cicdguy it is more difficult to explain than what you see in the commits (silly me I know)
> 
> Totally understand that the autocommits are painful.
> 
> Technically, yes, it is possible to run one workflow after the other, but I would rather have them run simultaneously and report the inconsistencies so you, as a developer, can take the action to correct one or both (or more) issues in a single commit. I care for fast developer feedback, and that is accomplished with parallel workflows. Running them sequentially is not great from a performance and quick DevEx perspective. WDYT?

You are right, we would lose speed and we do not want that! Thanks for the feedback ;)

PR #1198, opened 2024-02-23
Replace `strat` with `strata`

### LINKED ISSUE #1186: `strat` vs `strata` (vs `strata_var`)
There are few places where we use `strat` and `strata` (and also `strata_var`) which seems to be inconsistent. 

image

image

It might be working if `strat` is required and `strata` is provided due to R's ability of partial match but we should avoid it (there is a seperate PR aims to punish it). The other way around is not functional.

Please analyse if that's correct.

If to be corrected, note that it's potentially a breaking change so would be nice to give a lifecycle warning and create `strata` from `strat` (or vice versa).

This has downstream consequences in `chevron`, `tmc` and TLG-Cat - if changed please amend these as well.

edelarua:
# Pull Request

Fixes #1186

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,8 @@
 
 ### Miscellaneous
 * Added function `expect_snapshot_ggplot` to test setup file to process plot snapshot tests and allow plot dimensions to be set.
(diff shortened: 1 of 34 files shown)


ayogasekaram (approved):
lgtm! Thanks for working on this @edelarua

PR #1201, opened 2024-03-07
add snaps for bland-altman function

### LINKED ISSUE #1175: adding g_bland_altman test snapshot

ayogasekaram:
closes #1175


tests/testthat/test-bland-altman.R
```diff
@@ -59,3 +59,29 @@ testthat::test_that("s_bland_altman works with two vectors with NA element in ei
   )
   expect_identical(res, expect, tolerance = 1e-5)
 })
+
+
+testthat::test_that("s_bland_altman works with customized parameters", {
+  set.seed(1)
+  x <- rnorm(20)
+  y <- rnorm(20)
```
shajoezhu, line 67: hi @ayogasekaram , 20 numbers is not too much. i think we can include in the code to avoid using random number generations

tests/testthat/test-bland-altman.R
```diff
@@ -59,3 +59,29 @@ testthat::test_that("s_bland_altman works with two vectors with NA element in ei
...
+  result <- s_bland_altman(x, y, 0.9)
+  res <- testthat::expect_silent(result)
+  testthat::expect_snapshot(res)
+})
+
+
+testthat::test_that("g_lineplot works with default settings", {
+  set.seed(1)
+  x <- rnorm(20)
+  y <- rnorm(20)
```
shajoezhu, line 77: hi @ayogasekaram , 20 numbers is not too much. i think we can include in the code to avoid using random number generations

shajoezhu (changes requested):
use fixed numbers, avoid regenerating numbers

shajoezhu (changes requested):
and updates, after https://github.com/insightsengineering/tern/pull/1200/files merged

Melkiades (commented):
@ayogasekaram thanks for this! Just one note for me: we should skip graph testing on ci for now right? And also is it calling nestcolor? I think I forgot to enforce it when reviewing original PR

ayogasekaram:
@Melkiades great point. the `g_bland_altman` function only calls base colours - not the nestcolors package. The `with_options` is setup for this test like the other graphs so I think it should be okay but I can remove it if it fails the integration tests :)

shajoezhu (approved):
lgtm! Thanks @ayogasekaram

PR #1207, opened 2024-03-11
rm emmeans from dependencies

shajoezhu:
# Pull Request

Fixes #nnn


pawelru:
I think that as long as this package is actively used ([link](https://github.com/insightsengineering/tern/blob/fc62f2f8a3eb7e4e781e9b99d9b42a84c287215e/R/summarize_ancova.R#L77)) we cannot drop it just like that. R CMD CHECK will fail for a good reason.

PR #1210, opened 2024-03-15
Refactor `g_km`

### LINKED ISSUE #1167: Investigate feasibility of exclusively using ggplot2 for all plots - part II (g_km)
See original issue https://github.com/insightsengineering/tern/issues/1109

Currently all but two plotting functions exclusively use ggplot2. These two functions were implemented using grobs to avoid issues when combining plots with accompanying annotation tables.

Look into feasibility of converting the following functions to use only ggplot2 (i.e. geom_text) instead of grobs:

edelarua:
# Pull Request

Fixes #1167

I'm not sure if the `grob`-related functions are used in other places - if so we may not be able to deprecate them currently. If/when we remove all of these functions we should be able to remove the dependencies on packages grid, gridExtra, and gtable.


Melkiades:
> # Pull Request
> Fixes #1167
> 
> I'm not sure if the `grob`-related functions are used in other places - if so we may not be able to deprecate them currently. If/when we remove all of these functions we should be able to remove the dependencies on packages grid, gridExtra, and gtable.

Talking with Kartik yesterday, I saw that grob is used to produce reports in teal.report. It may be useful then to move towards removing this dep also there, as it is not maintained (right?). @kartikeyakirar fyi

DESCRIPTION
```diff
@@ -36,6 +36,7 @@ Imports:
     forcats (>= 1.0.0),
     formatters (>= 0.5.5),
     ggplot2 (>= 3.5.0),
+    ggsurvfit (>= 1.0.0),
```
Melkiades, line 39: This would be in place of 3 packages right?

DESCRIPTION
```diff
@@ -36,6 +36,7 @@ Imports:
...
     grid,
     gridExtra (>= 2.0.0),
     gtable (>= 0.3.0),
```
Melkiades, line 42: these packages

NEWS.md
```diff
@@ -18,6 +21,8 @@
 * Began deprecation of the named element `strat`, to be renamed to `strata`, within the `variables` argument to `h_rsp_to_logistic_variables`, `h_logistic_mult_cont_df`, `h_odds_ratio_subgroups_df`, `h_coxreg_mult_cont_df`, `h_coxph_subgroups_df`, `h_tbl_coxph_pairwise`, `extract_rsp_biomarkers`, `extract_rsp_subgroups`, `extract_survival_biomarkers`, and `extract_survival_subgroups`.
 * Began deprecation of the `strat` argument to `s_coxph_pairwise` and replaced it with the `strata` argument.
 * Adapt to argument renames introduced in `ggplot2` 3.5.0.
+* Began deprecation of no longer used Kaplan-Meier helper functions `h_data_plot`, `h_ggkm`, `h_decompose_gg`, `h_km_layout`, `h_grob_tbl_at_risk`, `h_grob_median_surv`, `h_grob_y_annot`, and `h_grob_coxph`.
+* Began deprecation of `grob`/`grid` related functions `stack_grobs`, `arrange_grobs`, `draw_grob`, `decorate_grob`, `split_text_grob`, `decorate_grob_factory`, and `decorate_grob_set` which are no longer used in `tern`.
```
Melkiades, line 25: So many lines of code ahah. Well done!!

DESCRIPTION
edelarua: Yes!

Melkiades: should we move them to suggests maybe?

Teninq:
Change the current layout:
1. Merge the two annotation bars into one.
2. Move the legend into the plot.

Example:
image

shajoezhu:
hi @edelarua , can you please place the legend within the plot (the request was to remove it from the current place, as the x-axis will get better alignment and more readable)

I will refine the annotation request as a separate issue. Thanks!

DESCRIPTION
pawelru, line 39: can you please add `pharmaverse/ggsurvfit` to the `Config/Needs/verdepcheck`? This would allow us to test against its devel branch in the max verdepcheck strategy. This is to test against upcoming changes.

edelarua: Will do!

tests/testthat/test-decorate_grob.R
```diff
@@ -1,9 +1,11 @@
+# All functions deprecated
+
 testthat::test_that("decorate_grob returns no warnings when creating an empty plot", {
   titles <- "Edgar Anderson's Iris Data"
   footnotes <- "The species are Iris setosa, versicolor, and virginica."
 
   grid::grid.newpage()
-  testthat::expect_silent(grid::grid.draw(
+  suppressWarnings(testthat::expect_warning(grid::grid.draw(
```
pawelru, line 8: `suppressWarnings()` right next to `expect_warning()` just looks odd :)
Why it's needed? What you wanted to do here?

DESCRIPTION
pawelru, line 39: This means a new package needs to be internally validated - it's not currently. It claims to have 100% coverage which is awesome. But I don't know about its dependencies. Looking briefly - `patchwork` is missing. So we have at least 2 submissions...
I don't want to stop the development - it's about to be one step ahead of what's after merge.

tests/testthat/test-h_km.R
```diff
@@ -131,70 +108,53 @@ testthat::test_that("h_tbl_coxph_pairwise estimates HR, CI and pvalue", {
...
-  variables <- list(tte = "AVAL", is_event = "is_event", arm = "ARMCD")
+# Deprecated Functions ----
 
-  testthat::expect_silent(
-    h_grob_coxph(df = df, variables = variables)
-  )
+# h_data_plot ----
+testthat::test_that("h_data_plot works as expected", {
+  data <- test_fit
+  testthat::expect_warning(result <- h_data_plot(data))
```
pawelru, line 116: if this is suppose to catch lifecycle warning then I think we should be using [`lifecycle:: expect_deprecated()`](https://lifecycle.r-lib.org/reference/expect_deprecated.html). Even though it's not a new functionality, I discovered this quite recently :P 
It should be more _accurate_ in a sense of checking against specific class of warnings as opposed to any kind of warnings

tests/testthat/test-decorate_grob.R
edelarua: As per testthat v3.0, `expect_warning()` only captures one warning at a time now (https://testthat.r-lib.org/reference/expect_error.html). 

For these functions, which are deprecated but also use deprecated functions internally, a deprecation warning is printed for the function called as well as every instance of a deprecated function call within that function. We can either suppress additional warnings or call `expect_warning()` consecutively for each warning.

E.g. For this first example the `expect_warning()` checks for the `decorate_grob` warning and then we can either use `suppressWarnings()` to ignore all other warnings, or use `expect_warning()` 2 more times to capture the other warnings.

I don't really care which method we use, but used `suppressWarnings` to reduce repetition. Do you have a preference?

shajoezhu (approved):
looks really great! thanks @edelarua

tests/testthat/test-g_km.R
```diff
@@ -136,29 +111,54 @@ testthat::test_that("annot_at_risk_title parameter works as expected", {
       annot_at_risk_title = TRUE
     )
   )
-  vdiffr::expect_doppelganger(title = "g_km_at_risk_title", fig = g_km_at_risk_title)
+  expect_snapshot_ggplot("g_km_at_risk_title", g_km_at_risk_title, width = 8, height = 5)
```
Melkiades, line 114: Silly questions: is it the same? i.e. is expect_snapshot_file etc doing the comparison? does it have tolerance?

tests/testthat/test-g_km.R
```diff
@@ -1,50 +1,46 @@
...
+      variables = variables
+    )
+  )
+  expect_snapshot_ggplot("g_km_default", g_km_default, width = 8, height = 5)
+})
 
-  withr::with_options(
+testthat::test_that("g_km works with title/footnotes and annotation", {
+  g_km_title_footer <- withr::with_options(
     opts_partial_match_old,
```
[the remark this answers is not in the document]
edelarua: It happens for `ggplot2` functions as well.

tests/testthat/test-g_lineplot.R
```diff
@@ -4,16 +4,12 @@ adlb$AVISIT <- droplevels(adlb$AVISIT)
...
 testthat::test_that("g_lineplot works with default settings", {
-  testthat::skip_if_not_installed("vdiffr")
-
   g_lineplot <- g_lineplot(adlb, adsl)
 
   expect_snapshot_ggplot(title = "g_lineplot", fig = g_lineplot, width = 10, height = 8)
 })
 
 testthat::test_that("g_lineplot works with custom settings and statistics table", {
-  testthat::skip_if_not_installed("vdiffr")
```
Melkiades, line 15: can we take it out fully then?

tests/testthat/test-g_km.R
edelarua: It does the same thing - I made `expect_snapshot_ggplot` so we could control the sizing of the plot in the snapshot, but otherwise mostly copied the `expect_doppelganger` internal code.

You can see the function here: https://github.com/insightsengineering/tern/blob/86bd2d28b5bf6197ea1585f14b8b58a4c5de0382/tests/testthat/setup.R#L25

tests/testthat/test-g_lineplot.R
edelarua: Yes! I removed all mentions of `vdiffr`.

NEWS.md
```diff
@@ -21,6 +24,8 @@
 * Removed all instances of deprecated parameters `time_unit_input`, `time_unit_output`, `na_level` and `indent_mod`.
 * Removed deprecated functions `summarize_vars`, `control_summarize_vars`, `a_compare`, `create_afun_summary`, `create_afun_compare`, and `summary_custom`.
 * Began deprecation of the `forest_grob`, `forest_dot_line`, `forest_viewport`, `vp_forest_table_part`, and `grid.forest` functions.
+* Began deprecation of no longer used Kaplan-Meier helper functions `h_ggkm`, `h_decompose_gg`, `h_km_layout`, `h_grob_tbl_at_risk`, `h_grob_median_surv`, `h_grob_y_annot`, and `h_grob_coxph`.
+* Began deprecation of `grob`/`grid` related functions `stack_grobs`, `arrange_grobs`, `draw_grob`, `decorate_grob`, `split_text_grob`, `decorate_grob_factory`, and `decorate_grob_set` which are no longer used in `tern`.
```
Melkiades, line 28: would add the deps changes

tests/testthat/test-g_km.R
Melkiades: I saw it! Did not expect (funnily enough) that `expect_doppelganger` was so direct ^^ but well done Emily in taking out that dep

PR #1216, opened 2024-03-25
fix verdepcheck

pawelru:


.github/workflows/scheduled.yaml
```diff
@@ -4,6 +4,9 @@ name: Scheduled 🕰️
 on:
   schedule:
     - cron: '45 3 * * 0'
+  push:
+    branches:
+      - fix_verdepcheck
```
pawelru, line 9: to be removed

R/utils_ggplot.R
```diff
@@ -165,7 +165,7 @@ df2gg <- function(df,
                   hline = TRUE,
                   bg_fill = NULL) {
   # convert NAs to text
-  df <- as.data.frame(apply(df, 1:2, tidyr::replace_na, replace = "NA"))
+  df <- as.data.frame(apply(df, 1:2, tidyr::replace_na, replace = NA))
```
Melkiades, line 168: I think this defies the comment. @edelarua is this right?

pawelru: I wanted to confirm it with @edelarua 
Here, we got an error about casting numeric to character

``` r
df <- data.frame(a = 1:5, b = 1:5)
df[1, 1] <- NA
apply(df, 1:2, tidyr::replace_na, replace = "NA")
#> Error in `vec_assign()`:
#> ! Can't convert `replace` <character> to match type of `data` <integer>.
```

<sup>Created on 2024-03-26 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

DESCRIPTION
```diff
@@ -30,7 +30,7 @@ Imports:
     broom (>= 0.5.4),
     car (>= 3.0-13),
     checkmate (>= 2.1.0),
-    cowplot (>= 0.7.0),
+    cowplot (>= 1.0.0),
```
pawelru, line 33: `hjust`, `vjust` arguments used are available since `1.0.0`

.github/workflows/scheduled.yaml
pawelru:
```suggestion
```

pawelru:
OK this is ready for a review. verdepcheck tests are green as can be seen here: https://github.com/insightsengineering/tern/actions/workflows/scheduled.yaml?query=branch%3Afix_verdepcheck

R/utils_ggplot.R
edelarua: Fixed by https://github.com/insightsengineering/tern/pull/1216/commits/b6e9660b212acc45c8f833e4337ac7cdb9ca7e1c. The `NA`s (and all other values) are converted to text without using `replace_na`.

edelarua (approved):
This should be good to go now!

Issue #1240, opened 2024-04-30
`max_width` of `decorate_grob()` seems to be incorrectly handled


pawelru:
Please have a look at the below example

```r
library(tern)
library(grid)

text <- "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."

plot(1)
grid.newpage()
grid.draw(
    decorate_grob(
        NULL,
        titles = text,
        width_titles = unit(1, "npc"),
        footnotes = text,
        width_footnotes = unit(1, "npc")
    )
)
```
![image](https://github.com/insightsengineering/tern/assets/12943682/06c32e7a-344d-4ca9-ba00-7128a28c1add)

This produces a plot with title being outside of the plot range. It looks to me that the splitting mechanism is incorrectly handling the max width parameter. The value of `unit(1, "npc")` should indicate max viewport width but not beyond that.

Melkiades:
@pawelru, there are margins that are correctly handled by the default `grid::unit(1, "npc") - grid::unit(1.5, "cm")`. I tried many options, and they seem to work fine. It is fighting a bit with rstudio when it comes to knowing what is the default width of viewport, but I think it is an rstudio issue. It works perfectly in base R and with files (@BFalquet). We can enhance the documentation, but I think for the general user, we can simply discourage the visual appreciation from rstudio and look at the file exports for these cases

pawelru:
The fact that it's okey-ish for the default value does not change the fact that it cannot properly handle other values (e.g. the one I used in my example - `unit(1, "npc")`).
There might be something with local viewport. I haven't checked create-export scenario (as opposed to create-draw one).

The longer story is that I somehow got into an email chain with the user complaining that this is not reproducible - one one machine it looked significantly different that on the other. That was for the default value you mentioned. But let's leave that user issue for a while and focus on the example i provided.

Melkiades:
The issue should be fixed. For rstudio users the viewport width is not fixable as the conversion is off: https://github.com/rstudio/rstudio/issues/8559

PR #1245, opened 2024-05-15
fix decorate grob vp widths for titles and footers when right margin is present

Melkiades:
#1240


Melkiades:

Melkiades:
Can someone please help me review this? @ayogasekaram @edelarua Thanks!!

edelarua (approved):
Lgtm! Thanks Davide!!

pawelru:
Hey Davide. I tried with the example from the issue but I still can't make it work properly. `width_titles = grid::unit(1, "npc")` should indicate full width without margins but apparently it's not like that...
image


Tested in RStudio for now. For some reason I struggle to test in VSCode but I will look into this next.

pawelru:
On VSCode it's ok... Let me dig more into this:
image

Melkiades:
@pawelru rstudio does not work superwell with viewport as far as I could see. Everything works for base R on my machine. Have not tried VS but my output in rstudio is identical to vs

pawelru:
You are right. Base R should be the main testing environment. Tested it there and it's good. Also tested your feature branch versus the main and I can notice that the issue is fixed now. I think it's good to go.
One question though. Is this already covered by any test? I can see edits in one test but it's about margins - not the width of tittle and/or footnote. If not - let's add a new one.

Melkiades:
> You are right. Base R should be the main testing environment. Tested it there and it's good. Also tested your feature branch versus the main and I can notice that the issue is fixed now. I think it's good to go. One question though. Is this already covered by any test? I can see edits in one test but it's about margins - not the width of tittle and/or footnote. If not - let's add a new one.

It is a valid point; I could not think of a proper test for it. I can use the width of the viewport as a test. It is what needs to be respected for the wrapping at the end. 

Btw the wrapping core is not keeping spaces and manual \n as far as I can see. We might want to use the one I wrote for {formatters} that keeps those. What do you think?

pawelru:
> Btw the wrapping core is not keeping spaces and manual \n as far as I can see. We might want to use the one I wrote for {formatters} that keeps those. What do you think?

This I don't know to be honest. I'm afraid that I'm not that deep in this to make a statement. 

I'm going to approve as I can see that the issue had been resolved. Please make a call about the test. If it can't be done then I guess we need to live with that.

pawelru (approved):

Melkiades:
> > Btw the wrapping core is not keeping spaces and manual \n as far as I can see. We might want to use the one I wrote for {formatters} that keeps those. What do you think?
> 
> This I don't know to be honest. I'm afraid that I'm not that deep in this to make a statement.
> 
> I'm going to approve as I can see that the issue had been resolved. Please make a call about the test. If it can't be done then I guess we need to live with that.

I added and closed an issue to track down my research. Testing for this issue can be a bit difficult considering the inconsistencies between different viewport page sizes' conversions

PR #1247, opened 2024-05-16
Bumps `{rmarkdown}` minimal version

averissimo:
# Pull Request

Part of https://github.com/insightsengineering/coredev-tasks/issues/546

Necessary bump to overcome a lack of binary on ppm snapshots that causes an error on minimum strategies for `scheduled` workflows.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,7 @@ Depends:
     rtables (>= 0.6.8)
 Imports:
     broom (>= 0.5.4),
-    car (>= 3.1-2),
+    car (>= 3.0-13),
     checkmate (>= 2.1.0),
     cowplot (>= 1.0.0),
     dplyr (>= 1.0.0),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,7 @@ Depends:
     rtables (>= 0.6.7)
 Imports:
     broom (>= 0.5.4),
-    car (>= 3.0-13),
+    car (>= 3.1-1),
     checkmate (>= 2.1.0),
     cowplot (>= 1.0.0),
     dplyr (>= 1.0.0),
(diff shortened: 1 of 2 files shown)


averissimo:

averissimo:
`lme4` (second level dependency) fails to build as the version it requires is more recent than the first-level dependency

Unfortunatelly, it's building from source as of now, so there's no way to avoid this problem

```
car_3.1-2 (2023-03-30)
| lme4_1.1-32 (2023-03-14)
```

This issue needs further research.

averissimo:
[Failure](https://github.com/insightsengineering/tern/actions/runs/9565341893) in `min_isolated` (see below) and `release` (expected until `{formatters}` is released)

Suggested action: Let's merge this after the release (#1253) with `min_isolated` failing.

Context:

The reason for that failure is related to the old release date of `{car}` which predates any of the "buildable" `{lme4}` versions.

In other words, when  that strategy tries to install all dependencies of `{car}` using a snapshot of its release date, it will always fail.

Other than adding a lot of customization to `{verdepcheck}` there is no way of solving this.

I suggest that we mark this failure as "acceptable" and wait for a new release of `{car}` @pawelru

DESCRIPTION
```diff
@@ -28,7 +28,7 @@ Depends:
     rtables (>= 0.6.8)
 Imports:
     broom (>= 0.5.4),
-    car (>= 3.0-13),
+    car (>= 3.1-2),
```
averissimo, line 31:
```suggestion
    car (>= 3.0-13),
```

averissimo:
Scheduled action is green fro this branch:

[Scheduled 🕰️](https://github.com/insightsengineering/tern/actions/workflows/scheduled.yaml)

pawelru (approved):
👍 thanks!

Issue #1236, opened 2024-04-25
[Bug]: line disappears in g_lineplot when only one datapoint per timepoint is present


BFalquet:
### What happened?

reproducible example

```
library(tern)
adsl <- tern_ex_adsl
adlb <- tern_ex_adlb %>% dplyr::filter(ANL01FL == "Y", PARAMCD == "ALT", AVISIT != "SCREENING")
adlb$AVISIT <- droplevels(adlb$AVISIT)
adlb <- dplyr::mutate(adlb, AVISIT = forcats::fct_reorder(AVISIT, AVISITN, min))

adlb <- adlb |>
  dplyr::filter(USUBJID == "AB12345-BRA-1-id-105")

adsl <- adsl |>
  dplyr::filter(USUBJID == "AB12345-BRA-1-id-105")

g_lineplot(adlb, adsl, subtitle = "Laboratory Test:")
```
![image](https://github.com/insightsengineering/tern/assets/64274616/3ce78fa7-380e-4e2d-a30c-3f50e20295c8)

whereas one would like something like:

![image](https://github.com/insightsengineering/tern/assets/64274616/74af8b2e-0bfe-43a9-86f1-0bbd5ae5d891)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

Issue #1235, opened 2024-04-25
[Feature Request]: control line type in g_lineplot


BFalquet:
### Feature description

the linetype in g_linplot is determined automatically. It would be good to be able to control it the same way as we do for colors.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

Issue #1081, opened 2023-10-11
Adjust the width of errorbar in g_lineplot()


qiliu1013:
### What happened?

It's good to have fexisibility to adjust the width of the errorbar in g_lineplot(). `width = 0.45` is too small if the x scale is large.

```
  if (!is.null(interval)) {
    p <- p +
      ggplot2::geom_errorbar(
        ggplot2::aes(ymin = .data[[whiskers[1]]], ymax = .data[[whiskers[max(1, length(whiskers))]]]),
        width = 0.45,
        position = position
      )
```

### sessionInfo()

```R
> packageVersion("tern")
[1] ‘0.9.1’
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #1261, opened 2024-06-27
Update of `g_lineplot` to have line type, error bar width, and avoid empty levels in

Melkiades:
Completes #1235 #1236 #1081 and https://github.com/insightsengineering/teal.modules.clinical/issues/1197

### DIFF
--- a/R/g_lineplot.R
+++ b/R/g_lineplot.R
@@ -71,6 +71,7 @@
 #' @param newpage `r lifecycle::badge("deprecated")` not used.
 #' @param col (`character`)\cr color(s).
 #' @param linetype (`character`)\cr line type(s).
+#' @param errorbar_width (`numeric(1)`)\cr width of the error bars.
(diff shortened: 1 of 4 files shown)


R/g_lineplot.R
```diff
@@ -70,6 +70,8 @@
 #' @param table_font_size (`numeric(1)`)\cr font size of the text in the table.
 #' @param newpage `r lifecycle::badge("deprecated")` not used.
 #' @param col (`character`)\cr color(s).
+#' @param linetype (`character`)\cr line type(s).
```
BFalquet, line 73: accepted values for linetype (e.g. `"99"`) is not very well documented in `ggplot2`. might be worth adding an example or extend a little bit the description.

BFalquet (approved):
Nice, little comment inside about doc. I ll make the corresponding update in chevron.

Issue #1254, opened 2024-06-18
[Bug]: decorate_grob_set fails when combining empty string and spaces in footnotes


BFalquet:
### What happened?

This is a funny one:

This fails: 
```
x <- chevron::run(chevron::mng01, chevron::syn_data, dataset = "adlb")
x <- lapply(x, ggplot2::ggplotGrob)

res <- tern::decorate_grob_set(
  x,
  titles = "title",
  footnotes = c("", "a a") # empty string and space
)
#> Error in if (grid::convertWidth(linewidth + gapwidth + width_i, grid::unitType(width), : missing value where TRUE/FALSE needed
```

This works
```
res <- tern::decorate_grob_set(
  x,
  titles = "title",
  footnotes = c(" ", "a a") # space and space
)

res <- tern::decorate_grob_set(
  x,
  titles = "title",
  footnotes = c("", "axa") # empty string and no space
)

res <- tern::decorate_grob_set(
  x,
  titles = "title",
  footnotes = c(" ", "axa") # blank space and no space
) 

res <- tern::decorate_grob_set(
  x,
  titles = "title",
  footnotes = c("a a", "") # blank space and empty string
)
```

The only condition where the function fails is when the first element of the vector is empty string and the other contains a blank space.

also when footnotes is `NULL`, the error is not very user friendly
```
x <- chevron::run(chevron::mng01, chevron::syn_data, dataset = "adlb")
x <- lapply(x, ggplot2::ggplotGrob)

res <- tern::decorate_grob_set(
  x,
  titles = "title",
  footnotes = NULL
)
#> Error in strsplit(text, " "): non-character argument
Created on 2024-06-18 with [reprex v2.1.0](https://reprex.tidyverse.org/)
```

Have fun!

### Relevant log output

```R
> packageVersion("tern")
[1] ‘0.9.4.9009’
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

PR #1274, opened 2024-07-31
Fix handling of `decorate_grob` for empty strings and NULL in titles and footers

Melkiades:
#1254

### DIFF
--- a/R/decorate_grob.R
+++ b/R/decorate_grob.R
@@ -319,7 +319,7 @@ split_text_grob <- function(text,
     ti == ""
   })
   if (any(text_is_empty)) {
-    text[text_is_empty] <- "\n"
+    text[text_is_empty] <- " "
   }
 
   if (!grid::is.unit(x)) x <- grid::unit(x, default.units)
--- a/R/decorate_grob.R
+++ b/R/decorate_grob.R
@@ -310,11 +310,15 @@ split_text_grob <- function(text,
   if ((length(text) == 1 && text == "") || length(text) == 0) {
     stop(substitute(text), " is NULL or an empty string. Please provide a valid text.")
   }
-  if(any(sapply(text, function(ti) {is.na(ti)}))) {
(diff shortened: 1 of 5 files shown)


R/decorate_grob.R
```diff
@@ -306,6 +306,18 @@ split_text_grob <- function(text,
                             name = NULL,
                             gp = grid::gpar(),
                             vp = NULL) {
+  # Silly checks for text
+  if ((length(text) == 1 && text == "") || length(text) == 0) {
```
Melkiades, line 310: @BFalquet is this to be disallowed? ore should it be only " "

R/decorate_grob.R
```diff
@@ -306,6 +306,18 @@ split_text_grob <- function(text,
...
+    stop(substitute(text), " is NULL or an empty string. Please provide a valid text.")
+  }
+  if(any(sapply(text, function(ti) {is.na(ti)}))) {
+    stop(substitute(text), " contains NA values. Please provide a valid text.")
+  }
+  text_is_empty <- sapply(text, function(ti) {ti == ""})
+  if(any(text_is_empty)) {
+    text[text_is_empty] <- "\n"
```
Melkiades, line 318: " " is it a good substitute? "\n" does double newline

Melkiades: I think ahah

R/decorate_grob.R
```diff
@@ -306,6 +306,22 @@ split_text_grob <- function(text,
...
+  if (any(sapply(text, function(ti) {
+    is.na(ti)
+  }))) {
+    stop(substitute(text), " contains NA values. Please provide a valid text.")
+  }
+  text_is_empty <- sapply(text, function(ti) {
+    ti == ""
+  })
+  if (any(text_is_empty)) {
+    text[text_is_empty] <- "\n"
```
Melkiades, line 322:
```suggestion
    text[text_is_empty] <- " "
```

PR #1280, opened 2024-08-07
Fix `g_km` bug when multiple estimates are made at max time

### LINKED ISSUE #1279: [Bug Report] `g_km` fails with multiple max time records
When multiple records have estimates at max time, `g_km` fails with error:

```
Error in yval == “Survival” && data$estimate[data$time == max_time2] > : length = 2’ in coercion to ‘logical(1)’
```

edelarua:
# Pull Request

Fixes #1279 

Bug reported on Overflow but not able to replicate it.

### DIFF
--- a/R/g_km.R
+++ b/R/g_km.R
@@ -378,8 +378,8 @@ g_km <- function(df,
 
     y_rng <- ylim[2] - ylim[1]
 
-    if (yval == "Survival" && data$estimate[data$time == max_time2] > ylim[1] + 0.09 * y_rng &&
-      data$estimate[data$time == max_time2] < ylim[1] + 0.5 * y_rng) { # nolint
+    if (yval == "Survival" && all(data$estimate[data$time == max_time2] > ylim[1] + 0.09 * y_rng) &&
+      all(data$estimate[data$time == max_time2] < ylim[1] + 0.5 * y_rng)) { # nolint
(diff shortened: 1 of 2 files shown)


R/g_km.R
```diff
@@ -378,8 +378,8 @@ g_km <- function(df,
 
     y_rng <- ylim[2] - ylim[1]
 
-    if (yval == "Survival" && data$estimate[data$time == max_time2] > ylim[1] + 0.09 * y_rng &&
-      data$estimate[data$time == max_time2] < ylim[1] + 0.5 * y_rng) { # nolint
+    if (yval == "Survival" && all(data$estimate[data$time == max_time2] > ylim[1] + 0.09 * y_rng) &&
+      all(data$estimate[data$time == max_time2] < ylim[1] + 0.5 * y_rng)) { # nolint
```
Melkiades, line 382: it seems a good fix! I think we need to ask the roverflow requester to test it

Melkiades (approved):
Unblocking this if the user does not add a review ;) Thanks Emily!!!

shajoezhu:
lets give this another day, i just pinged the user earlier today

shajoezhu:
hi @edelarua , this is good to go. user was happy with outcome. I think we need to add some test for this. something like snapshot, but not the picture itself, but ggplot datastucture.

PR #1321, opened 2024-10-08
Update DESCRIPTION

shajoezhu:
use survival 3.6-4

# Pull Request

Fixes #nnn

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,7 +48,7 @@ Imports:
     rlang (>= 1.1.0),
     scales (>= 1.2.0),
     stats,
-    survival (>= 3.7-0),
+    survival (>= 3.6-4),
     tibble (>= 2.0.0),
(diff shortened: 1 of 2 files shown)


edelarua:
Hi Joe,

Sorry about this - since https://github.com/therneau/survival/issues/240 has been fixed I thought we could remove the workarounds for the partial argument matching, but I forgot to check which validated version was available. I have reverted the removal of the partial argument matching code as well.

shajoezhu:
Hi @edelarua , thanks so much for the changes. Actually I think it was a good change, but maybe we did it too soon? 

I like your changes, and I think we should reinstate them at some stage. 

I have a proposal, as survival is a package is that is potentially used by other packages as well. I am going to submit the new version to the autovalidation, and see if it can pass the validation. once we have the new version, we can then safely implemtnat this.

Issue #1338, opened 2024-10-24
rhub - nosuggests fails


pawelru:
https://github.com/insightsengineering/tern/actions/runs/11423183175/job/31782077445

```
* checking examples ... ERROR
Running examples in ‘tern-Ex.R’ failed
The error most likely occurred in:
> ### Name: g_forest
> ### Title: Create a forest plot from an 'rtable'
> ### Aliases: g_forest
> 
> ### ** Examples
> 
> library(dplyr)
Attaching package: ‘dplyr’
The following objects are masked from ‘package:stats’:
    filter, lag
The following objects are masked from ‘package:base’:
    intersect, setdiff, setequal, union
> library(forcats)
> library(nestcolor)
Error in library(nestcolor) : there is no package called ‘nestcolor’
Execution halted
* checking tests ...
  Running ‘testthat.R’
 ERROR
```

```
── Failed tests ────────────────────────────────────────────────────────────────
  Error ('test-g_lineplot.R:7:3'): g_lineplot works with default settings
  <packageNotFoundError/error/condition>
  Error in `loadNamespace(x)`: there is no package called 'nestcolor'
  Backtrace:
       ▆
    1. ├─testthat::expect_silent(g_lineplot <- g_lineplot(adlb, adsl)) at test-g_lineplot.R:7:3
    2. │ └─testthat:::quasi_capture(enquo(object), NULL, evaluate_promise)
    3. │   ├─testthat (local) .capture(...)
    4. │   │ ├─withr::with_output_sink(...)
    5. │   │ │ └─base::force(code)
    6. │   │ ├─base::withCallingHandlers(...)
    7. │   │ └─base::withVisible(code)
    8. │   └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
    9. ├─tern::g_lineplot(adlb, adsl)
   10. └─base::loadNamespace(x) at tern/R/g_lineplot.R:413:3
   11.   └─base::withRestarts(stop(cond), retry_loadNamespace = function() NULL)
   12.     └─base (local) withOneRestart(expr, restarts[[1L]])
   13.       └─base (local) doWithOneRestart(return(expr), restart)
  
  Error ('test-g_lineplot.R:13:3'): g_lineplot works with custom settings and statistics table
  <packageNotFoundError/error/condition>
  Error in `loadNamespace(x)`: there is no package called 'nestcolor'
  Backtrace:
       ▆
[...]

PR #1339, opened 2024-10-24
Move nestcolor to imports

### LINKED ISSUE #1312: Move nestcolor from suggests to import
I don't really want to drop "nestcolor::theme_nest()" from the g_lineplot default option, reason being this is primarily used by roche, and the nest default values are set for a reason, 

I don't want to replace the usage of nestcolor either, feel like this is going backwards, and adding code that we have to maintain in tern ...

one possibility is to move nestcolor from suggests to import @pawelru @khatril , what do you guys think?

_Originally posted by @shajoezhu in https://github.com/insightsengineering/tern/issues/1293#issuecomment-2333168526_

edelarua:
# Pull Request

Fixes #1312 and #1338


Melkiades (approved):
Lgtm!! thanks Emily ;)

PR #1379, opened 2025-01-22
Fix label and indentation processors

### LINKED ISSUE #1378: Fix label and indentation processors
Upon further testing after merging in https://github.com/insightsengineering/tern/pull/1361 I noticed a few unintended changes to output caused by modifications made to `get_labels_from_stats()` and `get_indents_from_stats()`. These functions should be updated to handle both the old and new inputs.

edelarua:
# Pull Request

Fixes #1378 #1363


tests/testthat/_snaps/count_patients_with_flags.md
```diff
@@ -599,10 +599,10 @@
       RowsVerticalSection (in_rows) object print method:
       ----------------------------
                     row_name formatted_cell indent_mod    row_label
-      1   count_fraction.SER   128 (78.05%)          2   Serious AE
+      1   count_fraction.SER   128 (78.05%)          2    New label
```
shajoezhu, line 602: should this change?

edelarua: Yes that's the intended behaviour. It was overwritten by #1378 but this is the correct result

R/utils_default_stats_formats_labels.R
```diff
@@ -362,29 +364,46 @@ get_indents_from_stats <- function(stats, indents_in = NULL, row_nms = NULL) {
...
-  # Modify some with custom labels
+  # Modify with custom indentation
   if (!is.null(indents_in)) {
-    # Stats is the main
-    common_names <- intersect(names(out), names(indents_in))
-    out[common_names] <- indents_in[common_names]
+    if (is.null(row_nms)) { # One row per statistic
+      common_names <- intersect(names(out), names(indents_in))
+      out[common_names] <- indents_in[common_names]
+    } else if (!is.null(row_nms)) { # One row per combination of variable level and statistic
```
Melkiades, line 386: I think it would be better to merge behavior with `.adjust_stats_desc_by_in_def` (we can definitely pick a better name ahah)

Melkiades: This is something to do for formats and indentation. Are the count cases that provide counts for each level for each stats (I think keeping this structure clear with the stat name on top is the best)

R/utils_default_stats_formats_labels.R
```diff
@@ -309,30 +303,23 @@ get_labels_from_stats <- function(stats, labels_in = NULL, levels_per_stats = NU
...
-      common_names <- intersect(names(out), names(labels_in))
-      out[common_names] <- unlist(labels_in[common_names], recursive = FALSE)
-    }
-  }
+  # Set missing labels to stat name
+  no_label <- sapply(out, is.na) %>% suppressWarnings()
+  out[no_label] <- names(out[no_label])
 
-  out
+  out %>% unlist()
```
Melkiades, line 322: maybe we want the other .unlist_keep*?

R/count_values.R
```diff
@@ -188,18 +195,18 @@ count_values <- function(lyt,
                          ...,
                          table_names = vars,
                          .stats = "count_fraction",
+                         .stat_names_in = NULL,
```
Melkiades, line 198: I think it would be better to use .stat_names as user-facing parameter

R/count_patients_with_flags.R
```diff
@@ -129,43 +130,43 @@ a_count_patients_with_flags <- function(df,
...
+  x_stats <- x_stats[.stats]
 
-  # Ungroup statistics with values for each level of x
-  x_ungrp <- ungroup_stats(x_stats, .formats, list())
-  x_stats <- x_ungrp[["x"]] %>% setNames(x_nms)
-  .formats <- x_ungrp[[".formats"]] %>% setNames(x_nms)
+  # Unlist stats
+  x_stats <- x_stats %>%
+    .unlist_keep_nulls() %>%
+    setNames(names(.formats))
```
Melkiades, line 155: why names(.formats))?

R/count_patients_with_flags.R
```diff
@@ -129,43 +130,43 @@ a_count_patients_with_flags <- function(df,
...
-  x_nms <- paste(rep(.stats, each = length(flag_variables)), flag_variables, sep = ".")
-  new_lbls <- if (!is.null(.labels)) .labels[names(.labels) %in% x_nms] else NULL
-  .labels <- .unlist_keep_nulls(get_labels_from_stats(.stats, .labels,
-    levels_per_stats = lapply(x_stats, names)
-  )) %>%
-    setNames(x_nms)
-
-  # indent mod formatting
-  .indent_mods <- get_indents_from_stats(.stats, .indent_mods, row_nms = flag_variables)
+  levels_per_stats <- rep(list(names(flag_variables)), length(.stats)) %>% setNames(.stats)
```
Melkiades, line 142: why not `lapply(x_stats, names)`?

R/count_occurrences.R
```diff
@@ -175,26 +175,27 @@ a_count_occurrences <- function(df,
...
-  .labels <- .unlist_keep_nulls(get_labels_from_stats(.stats, .labels, levels_per_stats = lapply(x_stats, names)))
-  .indent_mods <- get_indents_from_stats(.stats, .indent_mods, row_nms = names(x_stats[[1]]))
-
   x_stats <- x_stats[.stats]
+  levels_per_stats <- lapply(x_stats, names)
+  .formats <- get_formats_from_stats(.stats, .formats, levels_per_stats)
+  .labels <- get_labels_from_stats(.stats, .labels, levels_per_stats)
+  .indent_mods <- get_indents_from_stats(
+    .stats, .indent_mods, levels_per_stats,
+    rep(0L, length(levels_per_stats[[1]])) %>% as.list() %>% setNames(levels_per_stats[[1]])
```
Melkiades, line 184: why is it different?

R/analyze_variables.R
```diff
@@ -602,44 +597,47 @@ a_summary <- function(x,
   )
 
   x_stats <- x_stats[.stats]
-  if (is.character(x) || is.factor(x)) {
-    levels_per_stats <- lapply(x_stats, names) # if there is a count is table() with levels
+
+  is_char <- is.character(x) || is.factor(x)
+  if (is_char) {
```
Melkiades, line 602: I think this and pval can still be simplified but we can do it in another PR ;)

R/count_values.R
edelarua: You're right, sorry. I think I just copy-pasted the wrong name here :) I'll update that!

R/count_patients_with_flags.R
edelarua: This isn't actually necessary since `.stats` is not required to have the same names as `.formats`, `.labels`, etc. so we can leave it as is with the default "`n.n`" type names. I was just testing some things when I put this in, but I'll remove it now!

R/count_patients_with_flags.R
edelarua: For this functions specifically, the names of `x_stats` correspond to the variable labels and not the variable names which is what we want to extract here.

i.e. we would want `levels_per_stats$count` to have elements `fl1`, `fl2`, `fl3`, not `Total AEs`, `Total number of patients with at least one adverse event`, `Total number of patients with fatal AEs`.

R/analyze_variables.R
edelarua: Yes, I agree!

R/count_occurrences.R
edelarua: Oh right, this is no longer needed since I changed the default `tern_defaults` for `get_indents_from_stats()`. I'll remove now!

R/count_patients_with_flags.R
Melkiades: I think your correction is appropriate. It was a bug afaik. The second "n" in that case could not be changed in any intuitive or clear way and it should change as it does now. Do you think we should allow a more complex edit as it happens for the other get_*? so if I want to change only one n I can? Or do we keep it simple for now?

Melkiades (approved):
Lgtm!!! Great great job Emily. Thanks

R/count_patients_with_flags.R
edelarua: New issue opened for this piece here: https://github.com/insightsengineering/tern/issues/1380

PR #1392, opened 2025-02-19
Retire `make_afun()` - Batch 2

### LINKED ISSUE #1382: Retire `make_afun()` - Batch 2
### Feature description

Refactor functions listed as batch 2 to no longer use `make_afun()`.

### Code of Conduct

### Contribution Guidelines

### Security Policy

edelarua:
# Pull Request

Fixes #1382

### DIFF
--- a/R/summarize_ancova.R
+++ b/R/summarize_ancova.R
@@ -111,12 +111,13 @@ h_ancova <- function(.var,
 s_ancova <- function(df,
                      .var,
                      .df_row,
-                     variables,
                      .ref_group,
                      .in_ref_col,
+                     variables,
                      conf_level,
                      interaction_y = FALSE,
-                     interaction_item = NULL) {
+                     interaction_item = NULL,
+                     ...) {
   emmeans_fit <- h_ancova(.var = .var, variables = variables, .df_row = .df_row, interaction_item = interaction_item)
 
   sum_fit <- summary(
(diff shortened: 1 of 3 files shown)


shajoezhu:
all looks good downstream in scda.test https://github.com/insightsengineering/scda.test/pull/187

R/summarize_ancova.R
```diff
@@ -298,23 +298,11 @@ summarize_ancova <- function(lyt,
                              ...,
                              show_labels = "visible",
                              table_names = vars,
-                             .stats = NULL,
+                             .stats = c("n", "lsmean", "lsmean_diff", "lsmean_diff_ci", "pval"),
                              .stat_names = NULL,
-                             .formats = list(
```
Melkiades: should we not keep the formats visible?

edelarua: We typically only list out formats/labels if they aren't the standard since they're less likely to be changed vs. `.stats` which we do list out. I don't feel the need to list out all `.formats`/`.labels` since I think it will just add clutter that's not relevant to most people, but I'm not opposed to doing so if you feel we need to. I was under the impression that that was the reason we added `tern_default_formats` (and similar) in the first place - to avoid the repetition/clutter :)

Melkiades: I see; it makes sense to add them only when they differ from the default. Alternatively, we could add the default here when the formats were specified in the `make_afun`. I guess we can decide from case to case if it is different and if it has many values

Melkiades (approved):
Lgtm! Why no custom stats for this one? is it because it uses a summary output? Anyway we can come back to it later. Thanks Emily!!

R/summarize_ancova.R
```diff
@@ -232,7 +237,7 @@ a_ancova <- function(df,
   )
 
   # Fill in formatting defaults
-  .stats <- get_stats("summarize_ancova", stats_in = .stats)
+  .stats <- c(get_stats("summarize_ancova", stats_in = .stats), names(custom_stat_functions))
```
Melkiades, line 240: This forces the order. I have a fix ready, I will add it in another PR ;)

PR #1393, opened 2025-02-20
refactor geom sd and fix custom stat order

### LINKED ISSUE #286: Refactor geometric mean computation
Current implementation is `exp(mean(log(x))))` negative values gives NAs, should we also consider the case of negative values with `prod(sign(x))`? 

See [wiki](https://en.wikipedia.org/wiki/Geometric_mean#:~:text=additionally%2C%20if-,negative%20values,-of%20the) for example

Melkiades:
Fixes #286

### DIFF
--- a/R/analyze_variables.R
+++ b/R/analyze_variables.R
@@ -238,6 +238,9 @@ s_summary.numeric <- function(x, control = control_analyze_vars(), ...) {
   geom_verbose <- args_list[["geom_verbose"]] %||% FALSE # Additional info if requested
   checkmate::assert_flag(geom_verbose)
   x_no_negative_vals <- x
+  if (identical(x_no_negative_vals, numeric())) {
+    x_no_negative_vals <- NA
+  }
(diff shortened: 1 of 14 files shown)


R/analyze_variables.R
```diff
@@ -234,10 +234,25 @@ s_summary.numeric <- function(x, control = control_analyze_vars(), ...) {
...
+  if (identical(x_no_negative_vals, numeric())) {
+    x_no_negative_vals <- NA
+  }
   x_no_negative_vals[x_no_negative_vals <= 0] <- NA
+  if (geom_verbose) {
+    if (any(x <= 0)) {
+      warning("Negative values were converted to NA for calculation of the geometric mean.")
+    }
+    if (all(is.na(x_no_negative_vals))) {
+      warning("All values are negative or NA. The geometric mean is NA.")
```
edelarua, line 250:
```suggestion
      warning("Since all values are negative or NA, the geometric mean is NA.")
```

edelarua (approved):
Good to go!

PR #1395, opened 2025-02-21
Remove lookup-refs

m7pr:
Part of https://github.com/insightsengineering/coredev-tasks/issues/609

From now on, we will provide development dependencies in 
```
Remotes: repo/project@branch
```
format, so it's explicitly visible in the DESCRIPTION file and can be handled by `pak::install`, `renv::install` and `remotes::install`.

With development dependencies specified in CJ Pipelines configuration, this connection was hidden, and it was hard to install the package from the main branch (or any other branch) locally from user's machine.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -26,10 +26,6 @@ jobs:
       REPO_GITHUB_TOKEN: ${{ secrets.REPO_GITHUB_TOKEN }}
     with:
       deps-installation-method: setup-r-dependencies
-      lookup-refs: |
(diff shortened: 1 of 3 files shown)


m7pr:

m7pr:
recheck

Melkiades (approved):
Fine with me! Thanks

llrs-roche (approved):
nestcolor is required on a lower version than available on CRAN, no sure if it depends on a feature recently released or not. But there is no need for Remotes and lookup-refs can be safely removed now.

Issue #1397, opened 2025-02-24
Refactor remaining analysis functions to align with new style


edelarua:
Following https://github.com/insightsengineering/tern/issues/1375, there are still a few functions which already did not use `make_afun()` that should be refactored to match the new style:

- [x] `afun_riskdiff()`
- [x] `count_occurrences()`
- [x] `count_occurrences_by_grade()`
- [x] `count_patients_with_event()`
- [x] `count_patients_with_flags()`
- [x] `count_values()`
- [x] `estimate_incidence_rate()`
- [x] `summarize_change()`
- [x] `summarize_patients_exposure_in_cols()`
- [x] `surv_time()`
- [x] `tabulate_rsp_subgroups()`
- [x] `tabulate_survival_subgroups()`

PR #1398, opened 2025-02-25
Refactor `tabulate_*()` functions and deprecate their helper functions

edelarua:
# Pull Request

Part of #1397


R/summarize_functions.R
```diff
@@ -5,7 +5,6 @@
 #'
 #' * [add_rowcounts()]
 #' * [estimate_multinomial_response()] (with [rtables::analyze()])
-#' * [h_tab_one_biomarker()] (probably to deprecate)
```
Melkiades, line 8: ahah we caught this early enough

R/response_subgroups.R
```diff
@@ -441,11 +487,6 @@ d_rsp_subgroups_colvars <- function(vars,
       varlabels,
       ci = paste0(100 * conf_level, "% CI")
     )
-
-    # The `lcl`` variable is just a placeholder available in the analysis data,
-    # it is not acutally used in the tabulation.
-    # Variables used in the tabulation are lcl and ucl, see `a_response_subgroups` for details.
-    colvars[colvars == "ci"] <- "lcl"
```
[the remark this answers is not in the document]
edelarua: I'm not sure why (probably a compatability issue at some point) but the `"ci"` statistic used to keep the name `"lcl"` when made by combining the `"lcl"` and `"ucl"` statistics.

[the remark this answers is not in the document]
Melkiades: It seems something that can and should be achieved by users before getting into the analysis... So well done in removing it. It was a bit of a unpredictable behavior I believe

R/response_subgroups.R
```diff
@@ -129,40 +129,69 @@ extract_rsp_subgroups <- function(variables,
...
+      paste(as.character(df$biomarker), as.character(df$subgroup), sep = ".")
+    }
+  } else {
+    make.unique(as.character(df$subgroup))
+  }
 
-  afun_lst <- Map(
-    function(stat, fmt, na_str) {
-      function(df, labelstr = "", ...) {
-        in_rows(
```
Melkiades, line 152: very undebuggable

R/response_subgroups.R
```diff
@@ -129,40 +129,69 @@ extract_rsp_subgroups <- function(variables,
...
+                                 .stats = NULL,
+                                 .stat_names = NULL,
+                                 .formats = NULL,
+                                 .labels = NULL,
+                                 .indent_mods = NULL) {
+  # Check for additional parameters to the statistics function
+  dots_extra_args <- list(...)
+  extra_afun_params <- retrieve_extra_afun_params(names(dots_extra_args$.additional_fun_parameters))
+  dots_extra_args$.additional_fun_parameters <- NULL
+  cur_stat <- extra_afun_params$.var %||% .stats
```
Melkiades, line 144: why not .stats?

R/response_subgroups.R
```diff
@@ -129,40 +129,69 @@ extract_rsp_subgroups <- function(variables,
...
+  # Auto format handling
+  .formats <- apply_auto_formatting(.formats, x_stats, extra_afun_params$.df_row, extra_afun_params$.var)
+
+  # Get and check statistical names
+  .stat_names <- get_stat_names(x_stats, .stat_names)
+
+  in_rows(
+    .list = x_stats %>% .unlist_keep_nulls(),
+    .formats = .formats,
+    .names = .nms,
```
Melkiades, line 190: ah we might made this external one day (it is then use to make the ard variable entry)

R/survival_duration_subgroups.R
```diff
@@ -225,14 +252,22 @@ tabulate_survival_subgroups <- function(lyt,
...
   checkmate::assert_true(any(c("n_tot", "n_tot_events") %in% vars))
   checkmate::assert_true(all(c("hr", "ci") %in% vars))
+  if ("pval" %in% vars && !"pval" %in% names(df$hr)) {
+    warning(
+      'The "pval" statistic has been selected but is not present in "df" so it will not be included in the output ',
+      'table. To include the "pval" statistic, please specify a p-value test when generating "df" via ',
+      'the "method" argument to `extract_survival_subgroups()`. If method = "cmh", strata must also be specified via ',
+      'the "variables" argument to `extract_survival_subgroups()`.'
+    )
+  }
```
Melkiades, line 270: Do you think this warning should be added also to the other tabulations? I feel like there should be a general check that extract_ and tabulate_ are accessing the same variables

R/survival_duration_subgroups.R
```diff
@@ -300,26 +344,26 @@ tabulate_survival_subgroups <- function(lyt,
   # Add columns from table_survtime (optional)
   if (length(colvars_survtime$vars) > 0) {
     lyt_survtime <- split_cols_by(lyt = lyt, var = "arm")
```
Melkiades, line 346: Was this actually a bug?

tests/testthat/_snaps/response_subgroups.md
```diff
@@ -226,7 +229,7 @@
       res
     Output
       $vars
-      [1] "n"     "n_rsp" "prop"  "n_tot" "or"    "lcl"   "pval" 
+      [1] "n"     "n_rsp" "prop"  "n_tot" "or"    "ci"    "pval"
```
Melkiades, line 232: This is the change we were talking about. Still struggling to understand the idea behind it

R/survival_duration_subgroups.R
```diff
@@ -153,40 +153,67 @@ extract_survival_subgroups <- function(variables,
...
+  cur_stat <- extra_afun_params$.var %||% .stats
+  var_lvls <- if ("biomarker" %in% names(dots_extra_args) && "biomarker" %in% names(df)) {
+    if ("overall" %in% names(dots_extra_args)) {
+      as.character(df$biomarker)
+    } else {
+      paste(as.character(df$biomarker), as.character(df$subgroup), sep = ".")
+    }
+  } else {
+    make.unique(as.character(df$subgroup))
+  }
```
Melkiades, line 177: Where does this come from? Sorry, I am missing where from...

R/survival_duration_subgroups.R
```diff
@@ -153,40 +153,67 @@ extract_survival_subgroups <- function(variables,
...
+  # Main statistics taken from df
+  x_stats <- as.list(df)
+
+  # Fill in formatting defaults
+  .stats <- get_stats("tabulate_survival_subgroups", stats_in = cur_stat)
+  levels_per_stats <- rep(list(var_lvls), length(.stats)) %>% setNames(.stats)
+  .formats <- get_formats_from_stats(.stats, .formats, levels_per_stats)
+  .labels <- get_labels_from_stats(
+    .stats, .labels, levels_per_stats,
+    tern_defaults = as.list(as.character(df$subgroup)) %>% setNames(var_lvls)
```
Melkiades, line 193: This is because it is a factor output right?

R/survival_duration_subgroups.R
```diff
@@ -153,40 +153,67 @@ extract_survival_subgroups <- function(variables,
...
-        )
-      }
-    },
-    stat = names(.formats),
-    fmt = .formats,
-    na_str = na_str
+  # if empty, return NA
+  if (nrow(df) == 0) {
+    return(in_rows(.list = list(NA) %>% stats::setNames(cur_stat)))
+  }
```
Melkiades, line 182: Do you think this could be also general?

Melkiades (commented):
Amazing work Emily!! I felt lost at moments between files ahah. Do you think we could move the deprecated functions into their connected main files? I would do the same with the tests with a `deprecated --------------` line. wdyt?

R/survival_duration_subgroups.R
edelarua: This is already in the other function - I'm not sure why it was only missing from this one

R/survival_duration_subgroups.R
edelarua: To clarify, this part is to correctly/uniquely identify the row name since we are operating across columns, not rows (I will comment the code as well):

```r
  # Uniquely name & label rows
  var_lvls <- if ("biomarker" %in% names(dots_extra_args) && "biomarker" %in% names(df)) {
    if ("overall" %in% names(dots_extra_args)) { # label rows for (nested) biomarker tables - e.g. "AGE", "BMRKR1"
      as.character(df$biomarker)
    } else { # data rows for (nested) biomarker tables - e.g. "AGE.LOW", "BMRKR1.Total Patients"
      paste(as.character(df$biomarker), as.character(df$subgroup), sep = ".")
    }
  } else { # data rows for non-biomarker tables - e.g. "Total Patients", "F", "M"
    make.unique(as.character(df$subgroup))
  }
```

R/survival_duration_subgroups.R
edelarua: I checked and there's very few functions where this scenario ever comes up - I think I added it to `count_occurrences()` but it wasn't useful elsewhere.

R/response_subgroups.R
edelarua: Since it uses a column-wise analyze function the variables are given via `vars` instead of `.stats` so `.stats` isn't the identifier of the current statistic here. 
We could supply `extra_args` during the `split_cols_by_multivar()` step instead and then use `.stats`, etc. from there, but that drops the labels from `.formats`, `.labels`, etc. which would then have to be re-added in so I don't think it's worth it.

Melkiades: I understand! Maybe we could make the variable a bit more verbose? or add a comment there to clarify briefly?

R/survival_duration_subgroups.R
Melkiades: Thank you so much!! Now it is very clear

edelarua:
> Amazing work Emily!! I felt lost at moments between files ahah. Do you think we could move the deprecated functions into their connected main files? I would do the same with the tests with a `deprecated --------------` line. wdyt?

I grouped the 3 deprecated functions (+ their tests) into the same file so it's more organized now.

edelarua:
@Melkiades fyi this is the fix for the failing integration test in chevron: https://github.com/insightsengineering/tern/pull/1398/commits/ad9e775d05fb6c4d8e32bb94cbf7be76e96f7da0

The `prune_table()` function behaves weirdly when `character` is mixed with `numeric` in the same row, so we need to use `numeric()` for blank spaces in the table instead of `character()`.

R/response_subgroups.R
edelarua: Updated variable name to `cur_col_stat` for clarity :)

Melkiades (approved):
Lgtm!!! Thanks Emily for the fast updates :)

Issue #1390, opened 2025-02-17
R CMD CHECK note


pawelru:
https://github.com/r-universe/pharmaverse/actions/runs/13267690610/job/37039563796
https://github.com/r-universe/pharmaverse/actions/runs/13267690610/job/37039564732

```
* checking DESCRIPTION meta-information ... NOTE
  Missing dependency on R >= 4.1.0 because package code uses the pipe
  |> or function shorthand \(...) syntax added in R 4.1.0.
  File(s) using such syntax:
    ‘incidence_rate.R’
```

Issue #1249, opened 2024-05-29
export `tern:::as_factor_keep_attributes`


Melkiades:
I got some user-feedback and it seems it may be an useful function

Melkiades:
I can export it in the current PR

PR #1403, opened 2025-02-28
Batch 1 initial part

Melkiades:
Fixes part of #1381 and #1249 and #1390

### DIFF
--- a/R/prop_diff.R
+++ b/R/prop_diff.R
@@ -64,7 +64,8 @@ s_proportion_diff <- function(df,
                                 "ha", "newcombe", "newcombecc",
                                 "strat_newcombe", "strat_newcombecc"
                               ),
-                              weights_method = "cmh") {
+                              weights_method = "cmh",
+                              ...) {
   method <- match.arg(method)
   if (is.null(variables$strata) && checkmate::test_subset(method, c("cmh", "strat_newcombe", "strat_newcombecc"))) {
     stop(paste(
@@ -72,7 +73,7 @@ s_proportion_diff <- function(df,
       "permitted. Please choose a different method."
     ))
   }
-  y <- list(diff = "", diff_ci = "")
+  y <- list(diff = character(), diff_ci = character())
 
   if (!.in_ref_col) {
     rsp <- c(.ref_group[[.var]], df[[.var]])
@@ -151,6 +152,7 @@ s_proportion_diff <- function(df,
 #' @examples
 #' a_proportion_diff(
 #'   df = subset(dta, grp == "A"),
+#'   .stats = c("diff"),
 #'   .var = "rsp",
 #'   .ref_group = subset(dta, grp == "B"),
 #'   .in_ref_col = FALSE,
@@ -159,11 +161,75 @@ s_proportion_diff <- function(df,
 #' )
 #'
 #' @export
-a_proportion_diff <- make_afun(
-  s_proportion_diff,
-  .formats = c(diff = "xx.x", diff_ci = "(xx.x, xx.x)"),
-  .indent_mods = c(diff = 0L, diff_ci = 1L)
-)
+a_proportion_diff <- function(df,
+                              ...,
+                              .stats = NULL,
+                              .stat_names = NULL,
+                              .formats = NULL,
+                              .labels = NULL,
+                              .indent_mods = NULL) {
+  dots_extra_args <- list(...)
+
+  # Check if there are user-defined functions
+  default_and_custom_stats_list <- .split_std_from_custom_stats(.stats)
+  .stats <- default_and_custom_stats_list$default_stats
+  custom_stat_functions <- default_and_custom_stats_list$custom_stats
+
+  # Adding automatically extra parameters to the statistic function (see ?rtables::additional_fun_params)
+  extra_afun_params <- retrieve_extra_afun_params(
+    names(dots_extra_args$.additional_fun_parameters)
+  )
+  dots_extra_args$.additional_fun_parameters <- NULL # After extraction we do not need them anymore
+
+  # Main statistical functions application
+  x_stats <- .apply_stat_functions(
+    default_stat_fnc = s_proportion_diff,
+    custom_stat_fnc_list = custom_stat_functions,
+    args_list = c(
+      df = list(df),
+      extra_afun_params,
+      dots_extra_args
+    )
+  )
+
+  # Fill in with stats defaults if needed
+  .stats <- c(
+    get_stats("estimate_proportion_diff", stats_in = .stats),
+    names(custom_stat_functions)
+  )
+
+  x_stats <- x_stats[.stats]
+
+  # Fill in formats/indents/labels with custom input and defaults
+  .formats <- get_formats_from_stats(.stats, .formats)
+  .indent_mods <- get_indents_from_stats(.stats, .indent_mods)
+  if (is.null(.labels)) {
+    .labels <- sapply(x_stats, attr, "label")
+    .labels <- .labels[nzchar(.labels)]
+  }
+  .labels <- get_labels_from_stats(.stats, .labels)
+
+  # Auto format handling
+  .formats <- apply_auto_formatting(
+    .formats,
+    x_stats,
+    extra_afun_params$.df_row,
+    extra_afun_params$.var
+  )
+
+  # Get and check statistical names from defaults
+  .stat_names <- get_stat_names(x_stats, .stat_names) # note is x_stats
+  .stat_names <- paste0(.stat_names, "_", dots_extra_args$method)
+
+  in_rows(
+    .list = x_stats,
+    .formats = .formats,
+    .names = names(.labels),
+    .stat_names = .stat_names,
+    .labels = .labels %>% .unlist_keep_nulls(),
+    .indent_mods = .indent_mods %>% .unlist_keep_nulls()
+  )
+}
 
 #' @describeIn prop_diff Layout-creating function which can take statistics function arguments
 #'   and additional format arguments. This function is a wrapper for [rtables::analyze()].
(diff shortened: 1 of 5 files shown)


Melkiades (commented):
@edelarua I think we can merge this, your PR and then refactor riskdiff if we deem it necessary. I think the behavior there is strange. I would prefer adding a subfunction to a_* functions to deal with that cases. I manage to get a weird hybrid on the other PR but it is better to have a clean slate with most recent updates

R/count_missed_doses.R
```diff
@@ -85,10 +77,80 @@ s_count_missed_doses <- function(x,
...
+  levels_per_stats <- lapply(x_stats, names)
+
+  # Fill in formats/indents/labels with custom input and defaults
+  .formats <- get_formats_from_stats(.stats, .formats, levels_per_stats)
+  .indent_mods <- get_indents_from_stats(.stats, .indent_mods, levels_per_stats)
+  if (is.null(.labels)) {
+    .labels <- sapply(.unlist_keep_nulls(x_stats), attr, "label")
+    .labels <- .labels[nzchar(.labels) & !sapply(.labels, is.null) & !is.na(.labels)]
+  }
+  .labels <- get_labels_from_stats(.stats, .labels, levels_per_stats)
```
edelarua, line 127: Could use these as defaults so that users can specify some labels but not all and still use the defaults (from the label attribute) for the rest.
```suggestion
  labels_def <- c(lapply(x_stats[["count_fraction"]], attr, "label"), tern_default_labels)
  .labels <- get_labels_from_stats(.stats, .labels, levels_per_stats, labels_def)
```

R/count_cumulative.R
```diff
@@ -144,10 +145,80 @@ s_count_cumulative <- function(x,
...
+  levels_per_stats <- lapply(x_stats, names)
+
+  # Fill in formats/indents/labels with custom input and defaults
+  .formats <- get_formats_from_stats(.stats, .formats, levels_per_stats)
+  .indent_mods <- get_indents_from_stats(.stats, .indent_mods, levels_per_stats)
+  if (is.null(.labels)) {
+    .labels <- sapply(.unlist_keep_nulls(x_stats), attr, "label")
+    .labels <- .labels[nzchar(.labels) & !sapply(.labels, is.null) & !is.na(.labels)]
+  }
+  .labels <- get_labels_from_stats(.stats, .labels, levels_per_stats)
```
edelarua, line 195: Again can use these as defaults for greater customizability.
```suggestion
  labels_def <- lapply(x_stats[["count_fraction"]], attr, "label")
  .labels <- get_labels_from_stats(.stats, .labels, levels_per_stats, labels_def)
```
```

**Review by edelarua — CHANGES_REQUESTED** [review]
Hi @Melkiades,

Great work!! I just have a few comments/suggestions but otherwise everything looks good :)

File: R/count_cumulative.R (hunk 2 above)
Comment (Melkiades) [reply R/count_cumulative.R:195]: The issue is that sometimes it is `""` the label attribute.

> get_labels_from_stats("count", NULL, levels_per_stats = list("count" = c("a", "b")), tern_defaults = list("count" = ""))
$count.a
[1] ""

$count.b
[1] ""

> get_labels_from_stats("count")
$count
[1] "count"

Comment (Melkiades) [reply R/count_cumulative.R:195]: Should we take only one x_stats out? I know the default is only one here but maybe it is a bit too stringent?

File: R/count_cumulative.R [hunk 3]
```diff
@@ -172,33 +243,50 @@ a_count_cumulative <- make_afun(
...
                              show_labels = "visible",
                              na_str = default_na_str(),
                              nested = TRUE,
-                             ...,
                              table_names = vars,
+                             ...,
+                             na_rm = TRUE,
+                             lower_tail = TRUE,
+                             include_eq = TRUE,
                              .stats = NULL,
```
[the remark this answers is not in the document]
Melkiades: Done!

R/count_cumulative.R
```diff
@@ -172,33 +243,50 @@ a_count_cumulative <- make_afun(
...
                              var_labels = vars,
                              show_labels = "visible",
                              na_str = default_na_str(),
                              nested = TRUE,
-                             ...,
                              table_names = vars,
+                             ...,
+                             na_rm = TRUE,
+                             lower_tail = TRUE,
+                             include_eq = TRUE,
```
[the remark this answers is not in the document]
Melkiades: I was thinking to use the `...` as separator between what goes into the analysis function and what does not. I have no issue in changing it as you say, to get users more aware of custom variables

R/count_missed_doses.R
Melkiades: In this way there are repetitions in labels_def right?

R/prop_diff.R
```diff
@@ -64,15 +64,16 @@ s_proportion_diff <- function(df,
...
+                              ...) {
   method <- match.arg(method)
   if (is.null(variables$strata) && checkmate::test_subset(method, c("cmh", "strat_newcombe", "strat_newcombecc"))) {
     stop(paste(
       "When performing an unstratified analysis, methods 'cmh', 'strat_newcombe', and 'strat_newcombecc' are not",
       "permitted. Please choose a different method."
     ))
   }
-  y <- list(diff = "", diff_ci = "")
+  y <- list(diff = character(), diff_ci = character())
```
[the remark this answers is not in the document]
Melkiades: Done! thanks

R/count_missed_doses.R
edelarua: There shouldn't be since we don't have defaults for variable levels - it will just add the level-specific labels (each corresponding to a number of missed doses) to the defaults so that the new ones are added but the "n" default label is kept

R/count_cumulative.R
edelarua: I would assume that there's a reason why the label is set to `""` sometimes - if not, we should probably just remove or change that default. What do you mean by "take only one x_stats out"? Also, isn't `count_fraction` the only statistic returned by `s_count_cumulative`?

Melkiades: If you wan to add your custom stats here you cannot due to the x_stats[["count_fraction"]]. It is the only stat returned at the moment, I am not sure it will always be like that!

Melkiades: I tried to make it more general. See other comments ;)

R/utils_default_stats_formats_labels.R
```diff
@@ -307,8 +310,20 @@ get_formats_from_stats <- function(stats,
...
+
+  # If labels_in is NULL, use label_attr_from_stats
+  if (is.null(labels_in)) {
+    labels_in <- label_attr_from_stats
+    labels_in <- label_attr_from_stats[
+      nzchar(label_attr_from_stats) &
+        !sapply(label_attr_from_stats, is.null) &
+        !is.na(label_attr_from_stats)
+    ]
+  }
```
Melkiades, line 325: This is the more general solution I could think of. There is an issue. It is not picking it up from the custom functions... I propose to leave it as is because anyway you can change it as you whish with .labels

tests/testthat/test-count_cumulative.R
```diff
@@ -128,3 +129,58 @@ testthat::test_that("count_cumulative works with denom argument specified", {
...
+      show_labels = "hidden",
+      .stats = c("count_fraction",
+        "my_output" = function(x, ...) {
+          out <- list("a" = 1, "b" = 3)
+          attr(out, "label") <- "stat_function decides"
+          out
+        },
+        "my_output2" = function(x, ...) {
+          out <- list("min-max" = c("min" = 1, "max" = 2))
+          attr(out, "label") <- "stat_function is great"
```
Melkiades, line 177: These are not picked up by the current get_labels_from_stats

R/utils_default_stats_formats_labels.R
```diff
@@ -278,7 +277,8 @@ get_formats_from_stats <- function(stats,
   out <- .fill_in_vals_by_stats(levels_per_stats, formats_in, tern_defaults)
 
   # Default to NULL if no format
-  out[names(out) == out] <- list(NULL)
+  weird_case_to_check <- unlist(out, use.names = FALSE) == unlist(levels_per_stats, use.names = FALSE)
+  out[names(out) == out | weird_case_to_check] <- list(NULL)
```
Melkiades, line 281: This happens for "out" in the tests. @edelarua do you know why it is getting that as format from `.fill_in_vals_by_stats`?

Melkiades: of course this fix is temporary! I think we can find a better solution in .fill_in_vals_by_stats

R/utils_default_stats_formats_labels.R
edelarua: Sounds good to me!

R/utils_default_stats_formats_labels.R
edelarua: I don't get an error if I remove the `weird_case_to_check` code so I'm not sure.

edelarua (approved):
This looks good to me! Just one typo I saw but otherwise whenever you're ready you can test on scda.test & then merge.

R/utils_default_stats_formats_labels.R
Melkiades: ahah ok let me double check ^^

Melkiades: for me it bugs:
```
[ FAIL 1 | WARN 0 | SKIP 0 | PASS 16 ]

── Error (test-count_cumulative.R:163:3): Testing label behavior when s_* forecasts label attributes ──
Error in `format_value(rawvalues(x), format = format, output = output, 
    na_str = na_str)`: Unknown format label: 'a'. Run `list_valid_format_labels()` to get a list of all available formats.
```

But we can fix this if it ever surfaces again or if we get there some day in the future

shajoezhu: let's have an issue to track this.

PR #1408, opened 2025-03-10
upversion r min version, and survial depednecies

### LINKED ISSUE #1366: [known defect with package dependency on survival package] [update when R dependency over 4.4.0]
this will be closed by https://github.com/insightsengineering/tern/pull/1365

However, this issue needs to be revisited when we can upversion to R (≥ 4.4.0), given Matrix dependency https://cran.r-project.org/web/packages/Matrix/index.html, which is required by the survival package 3.8-3, https://cran.r-project.org/web/packages/survival/index.html

breaking change was flagged by survival package author

```
I am running a large number of checks on packages that depend on survival.  The rationale for the changes is given below.  For these  6 packages it was noted that they make use of the phrase  survival::strata() inside a formula, which doesn't do what people think it does: ggRandomForests, insight, mfp2, rineq, simtrial, and tern.   You might want to take a further look at your tests.   The new version 3.6-2 is on github and should be on CRAN shortly.

Terry Therneau

The latest version of the survival package has an important addition.  In prior code the call
     coxph(Surv(time, status) ~ age + strata(inst), data=lung)
 could fail if a version of either Surv() or strata() existed elsewhere on the search path; the wrong function could be picked up.  This has now been rectified.  
 
 1. For the coxph, concordance, survcheck, survfit, and survdiff modeling functions, any Surv(), strata(), cluster(), psline() or tt() model terms will use the versions from the survival namespace, automatically.
 
 2. Any use of survival::Surv(), survival::strata(), survival::cluster(), survival::pspline() or survival::tt() will have the 'survival::' modifier removed before the formula is evaluated.
 
 3. An exception to 1 above are models with a function/variable conflict such as Surv(time, status) ~ x + strata + strata(group) (found in the Epi package) or Surv(time, status) ~ x + strata(strata) (found in epiDisplay); I simply haven't thought of an automatic way to pick strata() from survival and the strata variable locally without stepping on my own feet.  In that same formula Surv() will be correctly dealt with, however.
 
 4. The call component of the relevant coxph, survfit, etc result is not changed, it still documents what you actually typed.  (Updating it  breaks a half dozen of the 800+ CRAN packages which make use of survival).
 
 This has no bearing on any discussion of whether to use coxph or survival::coxph as the primary call.  
 
One motivation for this is that use of :: messes up formula processing, in particular the "specials" argument of model.frame, something the survival package depends on.  For instance, in versions prior to survival3.8-0 the following two calls lead to completely different models; in the second strata is not recognized as special.
     coxph(Surv(time, status) ~ age + strata(inst), data=lung)
     coxph(Surv(time, status) ~ age + survival::strata(inst), data = lung)
Under the new regime both now work properly. 
```

_Originally posted by @shajoezhu in https://github.com/insightsengineering/tern/issues/1364#issuecomment-2559076808_

shajoezhu:
# Pull Request

Fixes #1366


edelarua (approved):
Lgtm!

shajoezhu:
downstream look ok https://github.com/insightsengineering/chevron/pull/830#issuecomment-2712549774
https://github.com/insightsengineering/scda.test/pull/191

Issue #1425, opened 2025-06-13
[Feature Request]: export various internal functions for usage in JJ based sister package to nest-tern


iaugusty:
### Feature description

export various internal functions that we use in the development of JJ package junco

We'll add documentation, examples and tests for these functions.
labels_or_names
as_factor_keep_attributes
assert_df_with_variables
assert_df_with_factors
assert_proportion_value
s_proportion_diff
s_test_proportion_diff
check_diff_prop_ci
s_surv_timepoint
range_noinf
clogit_with_tryCatch
get_covariates

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

Melkiades:
> [@Melkiades](https://github.com/Melkiades) I have added the extra exported functions into a category within _pkgdown.yml, for this purpose I also have added a badge (all as stable) to each of them. Please let me know if you'd prefer other approach for these. The only function that I have not assigned a category is range_noinf. To which category would you classify this. Could you take a look into the branch prior to initiate the PR?

You could open the PR as a draft. All the above things look good. Just remember to add examples for each exported function/doc page if not present already. `range_noinf` should be within the formatting functions

PR #1426, opened 2025-06-17
1425 feature request export various internal functions for usage in jj based sister package to nest tern

iaugusty:
# Pull Request

addresses #1425, export several internal functions


iaugusty:
hi @shajoezhu, in NEWS.md, should I mention which functions are now exported, or would an overall sentence like 
* Exported various functions that were `internal` in prior release. 
under the Enhancements paragraph
be sufficient?

Melkiades:
> hi @shajoezhu, in NEWS.md, should I mention which functions are now exported, or would an overall sentence like
> 
> * Exported various functions that were `internal` in prior release.
>   under the Enhancements paragraph
>   be sufficient?
> 
> Or would you prefer that I add this to ### Miscellaneous, like was done for Converted `as_factor_keep_attributes()` to an exported function. And that I add a line per exported function.

I would add the specific names into miscellaneous as it was done for `as_factor_keep_attributes()`

Melkiades (commented):
For me, it is already good to go. Thanks!! Just fyi, the lintr error appears only now because lintr processing analyzes only changed files. It is only one line that has a `return()` that is not needed ;)

Issue #1440, opened 2025-11-03
[Bug]: Keeping NAs with `analyze_vars(na_rm = TRUE)` raise an error


llrs-roche:
### What happened?

One user complained that teal modules don't handle well NAs. After some exploration it was shown that `analyze_vars()` can remove NAs correctly and there was some problem on the modules (https://github.com/insightsengineering/teal.modules.clinical/issues/1437). Modules used `tern::df_explicit_na()` to show NAs. However, they use it because `analyze_vars()` doesn't show NAs when `na_rm = FALSE`, instead it raises an error.

See the last two table layouts were NAs are not shown on tables even if they are not removed (layouts `tern_w_NA*`) and an error is raised, while the table using just rtables shows there are NAs present (layouts `w_NA*`). 

``` r
library("tern")
#> Loading required package: rtables
#> ...
dta <- data.frame(risk = factor(c(NA, "High Risk", "High Risk","High Risk", "Low Risk", "High Risk", "Low Risk", "High Risk", "High Risk"), levels = c(NA, "High Risk", "Low Risk")),
                  arm = factor(c("A", "C", "C", "C", "C", "C", "C", "C", "C"),levels = c("A", "B", "C")),
                  random  = as.factor(sample(letters, 9)))
w_NA <- basic_table() |>
  split_cols_by(var = "arm") |>
  split_rows_by(var = "random") |>
  analyze("risk", afun = function(x){
    as.matrix(table(x, useNA = "always")
  )}, inclNAs = TRUE)
w_NA_wo_split_rows <- basic_table() |>
  split_cols_by(var = "arm") |>
  analyze("risk", afun = function(x){
    as.matrix(table(x, useNA = "always")
  )}, inclNAs = TRUE)
tern_w_NA <- basic_table() |>
  split_cols_by(var = "arm") |>
  split_rows_by(var = "random") |>
  analyze_vars(vars = "risk", na_rm = FALSE)
tern_w_NA_wo_split_rows <- basic_table() |>
  split_cols_by(var = "arm") |>
  analyze_vars(vars = "risk", na_rm = FALSE)

build_table(w_NA, df = dta)
#>             A         B         C   
#> ————————————————————————————————————
#> d                                   
#>   risk   0, 0, 0   0, 0, 0   1, 0, 0
#> i                                   
#>   risk   0, 0, 0   0, 0, 0   0, 1, 0
#> n                                   
#>   risk   0, 0, 1   0, 0, 0   0, 0, 0
[...]

PR #1441, opened 2025-11-04
Fix NAs behavior

Melkiades:
#1440 

If this is unified correctly may have impacts downstream

### DIFF
--- a/R/analyze_variables.R
+++ b/R/analyze_variables.R
@@ -676,7 +676,8 @@ a_summary <- function(x,
 #' @describeIn analyze_variables Layout-creating function which can take statistics function arguments
 #'   and additional format arguments. This function is a wrapper for [rtables::analyze()].
 #'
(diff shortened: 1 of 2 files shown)


tests/testthat/test-analyze_variables.R
```diff
@@ -64,7 +64,6 @@ testthat::test_that("s_summary works when factors have NA levels", {
 
 testthat::test_that("s_summary works with factors with NA values handled and correctly removes them by default", {
   x <- factor(c("Female", "Male", "Female", "Male", "Male", "Unknown", "Unknown", "Unknown", "Unknown", NA))
-  x <- explicit_na(x)
```
Melkiades: this was a bit weird. If it is explicit, I should not guess what chr label to trim anyway, right? (here it was guessing `"<Missing>"`)

shajoezhu: this is good now right

shajoezhu (approved):
lgtm!

Issue #1428, opened 2025-07-18
[Bug]: g_km: x-axis alignment for plot and table


somb2025:
### What happened?

Hello,
The axis for the plot and table are aligned when no xticks are specified as shown below:

````
library(dplyr)
library(tern)
library(ggplot2)

df <- tern_ex_adtte %>%
  filter(PARAMCD == "OS") %>%
  mutate(is_event = CNSR == 0)
variables <- list(tte = "AVAL", is_event = "is_event", arm = "ARMCD")

g_km(df = df, variables = variables)

````
<img width="1106" height="834" alt="Image" src="https://github.com/user-attachments/assets/2b340b54-8917-4e71-b5a7-ac97bd05d133" />

However, when the xticks are specified, the plot and table are no longer aligned. Please can you look into this?
```
g_km(df = df, variables = variables, xticks = c(1000, 2000))
```

<img width="1132" height="843" alt="Image" src="https://github.com/user-attachments/assets/dc958f43-e38e-47f4-9582-090b3a6d1e48" />

Thank you.

### sessionInfo()

```R

```

### Relevant log output

```R

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [x] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [x] I agree to follow this project's Security Policy.

PR #1459, opened 2025-12-17
Fix xaxis alignment for ggkm

Melkiades:
#1428


Melkiades:
For aligned plots, see crane version of this plot. At the moment this is out of scope. It needs more refactoring

PR #1473, opened 2026-05-06
Use plotlist argument in all cowplot::plot_grid() calls

### LINKED ISSUE #208: Use all `plot_grid()` with plotlist
After user inputs, I discovered that this can be done:

We should do it also in NEST1.0

If your plot has already been combined into a cowplot grid (e.g., using plot_grid()), I have some bad news: you practically can't extract the original data from the final cowplot object.

Why You Can't Extract It
When you pass a ggplot object to cowplot::plot_grid(), cowplot converts the plot into a grid graphical object (a grob) using the ggplotGrob() function under the hood.

During this conversion, the raw underlying dataframe ($data) is stripped away. The plot stops being a dynamic mathematical mapping of variables and is baked into a static collection of visual drawing instructions (lines, polygons, text, and colors). As the creator of cowplot (Claus Wilke) has noted regarding plot_grid(): "Once that has happened, you can't go back."

(Note: cowplot does have helper functions like get_plot_component() or get_panel(), but these extract the visual grob elements—like a legend or an axis—to redraw them, not the raw dataset.)

The Workaround
If you need to build a cowplot grid and access the underlying data later, you have to intercept the plots before they are converted into grobs. The best practice is to store your original ggplot objects in a list:

```r
library(ggplot2)
library(cowplot)

# 1. Create your individual plots
p1 <- ggplot(mtcars, aes(x = wt, y = mpg)) + geom_point()
p2 <- ggplot(mtcars, aes(x = hp, y = qsec)) + geom_point()

# 2. Store them in a list BEFORE combining
plot_list <- list(plot_1 = p1, plot_2 = p2)

# 3. Pass the list to plot_grid
my_grid <- plot_grid(plotlist = plot_list)
By doing it this way, you can just extract the data from your saved list whenever you need it using the standard ggplot extraction methods:

R
# Extract the original dataframe for the first plot
data_p1 <- plot_list$plot_1$data

```

Melkiades:
# Pull Request

Fixes insightsengineering/crane#208 (same change applied to tern)

All `cowplot::plot_grid()` calls now pass plots via `plotlist = list(...)` instead of positional `...` arguments. This preserves the original ggplot objects in a list, making it possible to extract underlying data after grid composition.

### Changed files
- `R/g_forest.R`
- `R/g_km.R`
- `R/g_lineplot.R`


Melkiades (commented):
Deprecation issues fail on main are fixed too

Melkiades:
This is out of scope at the moment. We will see this in the future.

PR #1474, opened 2026-05-06
Add additional lr_stat_df statistic

### LINKED ISSUE #158: [Feature Request]:  tern: Inventory of possible  forked tern functions - Survival analysis: Cox PH hazard ratio with CI
### Feature description

- File: `R\coxph_hr.R`
- Functions:
  - `s_coxph_hr()`
  - `a_coxph_hr()`
- Origin: Forked from `tern::s_coxph_pairwise()` (plus a formatted afun wrapper).
- Purpose/Notes:
  - Adds `lr_stat_df` (log-rank statistic with degrees of freedom) in addition to the usual stats.
  - Supports different p-value methods (`wald`, `log-rank`, `likelihood`)
  - `a_coxph_hr()` formats the stats for `rtables`
- Exported: Yes

munoztd0:
# Pull Request

Purpose:

- Adds lr_stat_df (log-rank statistic with degrees of freedom) in addition to the usual stats.
- Adds alternative = c("two.sided", "less", "greater") logic

Fixes https://github.com/johnsonandjohnson/junco/issues/158


munoztd0:
@danielinteractive see https://github.com/johnsonandjohnson/junco/issues/158

Melkiades:
Hey, thanks for working on this! A few thoughts:

**`lr_stat_df` not registered in defaults** — The new statistic is returned by `s_coxph_pairwise()` but isn't added to `tern_default_stats` or `tern_default_formats` in `utils_default_stats_formats_labels.R` (line 577). Without that, it won't show up in tables unless the user explicitly requests it via `.stats`, and even then it won't have a format. See how `hr_ci_3d` is handled for reference — it's in the return list and has a format entry even though it's not in the defaults.

**`alternative` placement** — In tern, the existing precedent for `alternative` is in `s_test_proportion_diff()` (in `prop_diff_test.R`), where it's passed directly to the underlying stats functions (`prop.test`, `fisher.test`, `mantelhaen.test`) that handle one-sided logic natively. Here, the p-value halving is done manually (`pval / 2` or `1 - pval / 2`), which is a different pattern. A couple of concerns:

- The halving approach works for Wald and likelihood ratio tests, but for the log-rank test from `survdiff()` the chi-squared statistic is inherently non-directional. A proper one-sided log-rank test would use the signed Z-statistic (`sign(HR - 1) * sqrt(chisq)`), not just halving the two-sided p-value. The approximation is common in practice but worth documenting if intentional.
- The direction check uses the Cox model HR (`sum_cox$conf.int[1, 1] < 1`) regardless of `pval_method`, which mixes the Cox estimate with the log-rank statistic.
- It might also be worth considering whether `alternative` should live in `control_coxph()` to keep the API consistent — that's where `pval_method`, `ties`, and `conf_level` already are.

**Runtime assertion** — The `checkmate::assert_true(all.equal(log_rank_pvalue, original_survdiff$pvalue))` line will crash the function if there's a floating-point rounding difference. This could happen with edge-case data. If the goal is to validate the d.f. calculation, a test assertion (`expect_equal` with tolerance) would be safer than a runtime check.

**d.f. calculation** — The manual recomputation of degrees of freedom from `original_survdiff$exp` duplicates internal logic from `survival::survdiff`. For a two-group pairwise comparison (which is what `s_coxph_pairwise` always does — it compares `.ref_group` vs `df`), the d.f. is always 1. A simpler approach would be `c(original_survdiff$chisq, length(original_survdiff$n) - 1)`.

**Minor things:**
- The typo fix (`orginal` → `original`) is a nice catch 👍
- The NEWS entry combines two changes in one bullet — tern convention is one bullet per logical change.
- The extra `expect_warning` wrappers in the NA test (2 → 4) could use a comment explaining what the additional warnings are.
- The snapshot diff shows trailing whitespace/alignment changes unrelated to the feature.

danielinteractive:
Thanks @Melkiades for your review, I have helped with some stats related changes here - please have a look:

- simplified d.f. calculation
- removed runtime assertion
- used signed LR statistic for one-sided p-value

If that is fine then @munoztd0 can take care of other changes afterwards.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # tern 0.9.10.9006
 
+* Added `alternative` argument to `s_coxph_pairwise()` to allow one-sided hypothesis testing and added `lr_stat_df` to the parameters return list
```
Melkiades: These are 2 enhancements and should be under ### Enhancements in two separate bullet points

R/survival_coxph_pairwise.R
```diff
@@ -47,6 +47,7 @@ s_coxph_pairwise <- function(df,
                              strata = NULL,
                              strat = lifecycle::deprecated(),
                              control = control_coxph(),
+                             alternative = c("two.sided", "less", "greater"),
```
Melkiades: alternative should be part of control_coxph as per API design

R/survival_coxph_pairwise.R
```diff
@@ -99,19 +103,36 @@ s_coxph_pairwise <- function(df,
...
+      stats::pnorm(signed_lr_stat)
+    } else {
+      stats::pnorm(signed_lr_stat, lower.tail = FALSE)
+    }
+  }
+
+
   list(
     pvalue = formatters::with_label(unname(pval), paste0("p-value (", pval_method, ")")),
+    lr_stat_df = formatters::with_label(unname(c(log_rank_stat, log_rank_df)), "Log-rank Degrees of freedom"),
```
Melkiades, line 154: `lr_stat_df` is still not registered in `tern_default_stats` or `tern_default_formats` in `utils_default_stats_formats_labels.R`. It won't appear in tables without explicit .stats selection, and has no format.

Melkiades: There is a section for this statistic function

R/survival_coxph_pairwise.R
```diff
@@ -99,19 +103,36 @@ s_coxph_pairwise <- function(df,
...
+  log_rank_pvalue <- original_survdiff$pvalue
 
   pval <- switch(pval_method,
     "wald" = sum_cox$waldtest["pvalue"],
     "log-rank" = log_rank_pvalue, # pvalue from original log-rank test survival::survdiff()
     "likelihood" = sum_cox$logtest["pvalue"]
   )
+
+  # Handle one-sided alternatives.
+  if (alternative != "two.sided") {
```
Melkiades: The alternative argument applies the one-sided logic using the log-rank signed Z-statistic regardless of pval_method. If the user selects pval_method = "wald" or "likelihood", they still get a one-sided p-value derived from the log-rank statistic, not from the Wald or likelihood ratio test. That's inconsistent — the two-sided path respects pval_method, but the one-sided path ignores it.

Melkiades:
> Thanks @Melkiades for your review, I have helped with some stats related changes here - please have a look:
> 
> * simplified d.f. calculation
> * removed runtime assertion
> * used signed LR statistic for one-sided p-value
> 
> If that is fine then @munoztd0 can take care of other changes afterwards.

Thanks Daniel and @munoztd0!! Great work. I have only a couple of minor issues and one question about the pvalue ^^

R/survival_coxph_pairwise.R
danielinteractive: oh yeah very good point, thanks...

danielinteractive:
Thanks @Melkiades ! 
@munoztd0 can you take care of the remaining comments?

R/survival_coxph_pairwise.R
danielinteractive: For Wald we can do one-sided, for likelihood ratio test we really can't, I adjusted accordingly now

munoztd0:
> Thanks @Melkiades ! @munoztd0 can you take care of the remaining comments?

Yes I will do ! thanks for the rest

munoztd0:
> Hey, thanks for working on this! A few thoughts:
> 
> **`lr_stat_df` not registered in defaults** — The new statistic is returned by `s_coxph_pairwise()` but isn't added to `tern_default_stats` or `tern_default_formats` in `utils_default_stats_formats_labels.R` (line 577). Without that, it won't show up in tables unless the user explicitly requests it via `.stats`, and even then it won't have a format. See how `hr_ci_3d` is handled for reference — it's in the return list and has a format entry even though it's not in the defaults.
> 
> **`alternative` placement** — In tern, the existing precedent for `alternative` is in `s_test_proportion_diff()` (in `prop_diff_test.R`), where it's passed directly to the underlying stats functions (`prop.test`, `fisher.test`, `mantelhaen.test`) that handle one-sided logic natively. Here, the p-value halving is done manually (`pval / 2` or `1 - pval / 2`), which is a different pattern. A couple of concerns:
> 
> * The halving approach works for Wald and likelihood ratio tests, but for the log-rank test from `survdiff()` the chi-squared statistic is inherently non-directional. A proper one-sided log-rank test would use the signed Z-statistic (`sign(HR - 1) * sqrt(chisq)`), not just halving the two-sided p-value. The approximation is common in practice but worth documenting if intentional.
> * The direction check uses the Cox model HR (`sum_cox$conf.int[1, 1] < 1`) regardless of `pval_method`, which mixes the Cox estimate with the log-rank statistic.
> * It might also be worth considering whether `alternative` should live in `control_coxph()` to keep the API consistent — that's where `pval_method`, `ties`, and `conf_level` already are.
> 
> **Runtime assertion** — The `checkmate::assert_true(all.equal(log_rank_pvalue, original_survdiff$pvalue))` line will crash the function if there's a floating-point rounding difference. This could happen with edge-case data. If the goal is to validate the d.f. calculation, a test assertion (`expect_equal` with tolerance) would be safer than a runtime check.
> 
> **d.f. calculation** — The manual recomputation of degrees of freedom from `original_survdiff$exp` duplicates internal logic from `survival::survdiff`. For a two-group pairwise comparison (which is what `s_coxph_pairwise` always does — it compares `.ref_group` vs `df`), the d.f. is always 1. A simpler approach would be `c(original_survdiff$chisq, length(original_survdiff$n) - 1)`.
> 
> **Minor things:**
> 
> * The typo fix (`orginal` → `original`) is a nice catch 👍
> * The NEWS entry combines two changes in one bullet — tern convention is one bullet per logical change.
> * The extra `expect_warning` wrappers in the NA test (2 → 4) could use a comment explaining what the additional warnings are.
> * The snapshot diff shows trailing whitespace/alignment changes unrelated to the feature.

@Melkiades thanks for your review and comments

I have tackled all remaining issues I believe

except :
> * The snapshot diff shows trailing whitespace/alignment changes unrelated to the feature.

because I believe that if you are indeed talking about those whitespace there are related to the feature which adds p-value and make width wider 

image

Other than that I have fixed:
> * The NEWS entry combines two changes in one bullet — tern convention is one bullet per logical change.
> **`lr_stat_df` not registered in defaults**

@danielinteractive has fixed everything else I believe

Melkiades (approved):
Amazing work! Thank you both for acting fast and precisely! Good to go for me :)

munoztd0:
> Amazing work! Thank you both for acting fast and precisely! Good to go for me :)

Sorry, I tried to rerun the checks because they were all cancelled but re triggered the workflow approval also @Melkiades

PR #1481, opened 2026-06-02
Kaplan meier;  Added `range_with_cens_info` statistic to `s_surv_time()`

### LINKED ISSUE #160: [Feature Request]: tern: Survival time analysis: Kaplan-Meier summary
### Feature description

- File: `R\kaplan_meier.R`
- Functions:
  - `s_kaplan_meier()`
  - `a_kaplan_meier()`
- Origin: Forked from the `tern` package file `survival_time.R`
- Purpose/Notes:
  - Additional statistics beyond upstream: `quantiles_lower`, `quantiles_upper`, and `range_with_cens_info`
  - `a_kaplan_meier()` provides formatted output for `rtables`
- Exported: Yes (`a_kaplan_meier()`); `s_kaplan_meier()` is internal in docs but used by the afun

munoztd0:
# Pull Request
Fixes https://github.com/johnsonandjohnson/junco/issues/160

## Motivation
These enhancements have been in production use in the `junco` package
(https://github.com/johnsonandjohnson/junco) and are being contributed
back to tern.

## Changes
* Added `range_with_cens_info` statistic to `s_surv_time()`.
* Converted  `s_surv_time()`to exported functions.

These are all NULL default changes and should not cause any breaking changes


munoztd0:
@Melkiades there is unrelated lint issue to my PR, want me to check it out or let it be ?

Melkiades:
> @Melkiades there is unrelated lint issue to my PR, want me to check it out or let it be ?

it is not a blocker in general (you could merge the PR even if with those) but it is happening everytime we touch a file in {tern} so probably better to do correct it when it happens (it is only removing return calls where not needed to be explicit)

R/formatting_functions.R
```diff
@@ -573,3 +573,26 @@ apply_auto_formatting <- function(.formats, x_stats, .df_row, .var) {
...
+#' @param ... not used. Required for `rtables` interface.
+#'
+#' @return A string in the format `"min to max"`, with `+` appended to `min` and/or `max`
+#'   when the corresponding censoring flag is non-zero.
+#'
+#'
+#' @keywords internal
+format_range_cens <- function(x, ...) {
+  checkmate::assert_numeric(x, len = 4)
+  lo <- paste0(round(x[1], 1), if (x[3] != 0) "+")
```
Melkiades: Minor: The format_range_cens function uses if (x[3] != 0) which returns NULL when false (since there's no else). This works because paste0("5", NULL) returns "5", but it's relying on an implicit R behavior that's not obvious. An explicit if/else with "" would be clearer:

`lo <- paste0(round(x[1], 1), if (x[3] != 0) "+" else "")`

R/survival_time.R
```diff
@@ -116,6 +121,10 @@ s_surv_time <- function(df,
     ),
     quantiles_upper = formatters::with_label(
       unname(srv_qt_ci[[2]]), paste0(quantiles[2] * 100, "%-ile (", f_conf_level(conf_level), ")")
+    ),
+    range_with_cens_info = formatters::with_label(
+      c(range, lower_censored, upper_censored),
+      "Min, max"
```
Melkiades: The existing range stat uses "Min - Max". The new stat uses "Min, max" (lowercase, comma). This inconsistency will look odd when both appear in the same table. Consider "Range (with censoring)" or "Min - Max (+)" to distinguish it from range while keeping consistent casing.

R/formatting_functions.R
```diff
@@ -573,3 +573,26 @@ apply_auto_formatting <- function(.formats, x_stats, .df_row, .var) {
...
+  hi <- paste0(round(x[2], 1), if (x[4] != 0) "+")
+  paste(lo, "to", hi)
```
Melkiades: The other range formats (range, range_censor, range_event) use "xx.x to xx.x" or "xx.x - xx.x" format strings, which respect the rtables formatting pipeline. Hardcoding round(x, 1) in a custom format function bypasses that and won't respect any user-specified precision. Consider using the same "xx.x to xx.x" format string approach and handling the + suffix differently, or at least making the decimal places configurable.

R/survival_time.R
```diff
@@ -52,7 +52,7 @@ NULL
 #'   * `range_event`: Survival time range for observations with events.
 #'   * `range`: Survival time range for all observations.
 #'
```
Melkiades, line 55: missing description

R/formatting_functions.R
danielinteractive: Sure can do... just my 2 cents that it is pretty clear and documented, see e.g. in the `if` documentation in R: "if returns the value of the expression evaluated, or NULL invisibly if none was (which may happen if there is no else)."
😉

R/survival_time.R
danielinteractive: @munoztd0 I think fine to change, then we just adjust the table scripts accordingly with custom label

R/formatting_functions.R
danielinteractive: Not sure if we can handle the `+` suffix differently. So might need to have an argument for decimal places then. (Unless we drill into the `formatters` pipeline or wherever the format pipeline logic lives but then it will be a lot of work?)

R/formatting_functions.R
munoztd0: Nice to know, changed to if else

R/survival_time.R
```diff
@@ -124,7 +124,7 @@ s_surv_time <- function(df,
     ),
     range_with_cens_info = formatters::with_label(
       c(range, lower_censored, upper_censored),
-      "Min, max"
+      "Min - Max"
```
Melkiades: I think this is even more confusing as identical to the other range. The option is "Min - Max (+)"

R/survival_time.R
munoztd0: done

R/survival_time.R
Melkiades: or "Min - Max (with censoring)"

R/survival_time.R
munoztd0: added

R/survival_time.R
munoztd0: waiting for @danielinteractive take on this one

danielinteractive: I think it is ok to take the suggestion because we will anyway need to change to our preferred custom label in the template scripts now

munoztd0: ok so I changed to "Min - Max (with censoring)" as per suggested

R/formatting_functions.R
munoztd0: I took a try at it to add configurable precision (also added some examples and tests and refactored it to match the other similar formatting functions).
@danielinteractive let me know what you think of it

R/formatting_functions.R
munoztd0, line 1: this is just to make the lintr happy

tests/testthat/_snaps/survival_time.md
```diff
@@ -107,16 +117,28 @@
...
-      1          median                  24.8          0           Median
-      2       median_ci        (21.10, 31.35)          0    Median 95% CI
-      3    median_ci_3d 24.76 (21.10 - 31.35)          0  Median (95% CI)
-      4       quantiles           10.8 - 47.6          0  25% and 75%-ile
-      5 quantiles_lower  10.81 (6.65 - 13.43)          0 25%-ile (95% CI)
-      6 quantiles_upper 47.60 (39.27 - 57.82)          0 75%-ile (95% CI)
-      7    range_censor           0.8 to 78.9          0 Range (censored)
-      8     range_event          0.1 to 155.5          0    Range (event)
-      9           range           0.1 - 155.5          0        Min - Max
+                     row_name        formatted_cell indent_mod
```
munoztd0, line 120: the new  "Min - Max (with censoring" made the snapshot diff be bigger but it is still just a one liner change functionally

Melkiades: it is fine! thanks for the modification

R/formatting_functions.R
danielinteractive: yeah I think that would be ok for me... if you want to make it even cooler you could try to take a "x.xx" kind of string and then use that for formatting the numbers internally

munoztd0:
@Melkiades I think I addressed all previous comments

Melkiades (approved):
Lgtm!! Thanks @munoztd0 and @danielinteractive for your work!!

munoztd0:
> Lgtm!! Thanks @munoztd0 and @danielinteractive for your work!!

Of course , I am glad it got back into tern.

Can we merge the PRs now ?

PR #1483, opened 2026-06-15
1482: Add CMH test with Sato variance estimator

### LINKED ISSUE #1482: [Feature Request]: Add `cmh_sato` also for `prop_diff_test`
### Feature description

Currently we just have it for the estimation, but we also need it for testing proportion differences 😄 

It is not trivial because we need to calculate the chi-squared statistic with the Sato variance estimator.

- Can we basically just use the CI calculation path and then take (diff/se)^2 as the statistic? 
  - checked: yes, this will work  
- And can we still apply the WH transformation on that too? i.e. are these two method options orthogonal to each other?
  - not clear. probably better not allow that - the user interface will only allow one and we can also assert this in `prop_cmh()`

A couple of expected values under given scenarios:

- 21/63 vs. 11/63, with 4 strata:
  - 5/16, 4/10, 7/20, 5/17 vs. 2/16, 3/11, 2/19, 4/17  
  - should give p = 0.035 
- 25/62 vs. 11/63 overall, with 4 strata:
  - 7/15, 3/10, 10/20, 5/17 vs. 2/16, 3/11, 2/19, 4/17  
  - should give p = 0.004

To do:

- [x] Add test cases from above for `cmh_sato` prop diff test calls
- [x] Implement a common helper that does the `cmh_sato` style calc and can be used both for CI and test funs
- [x] Refactor CI fun accordingly
- [x] Refactor test fun accordingly

### Code of Conduct

### Contribution Guidelines

### Security Policy

danielinteractive:
closes #1482

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,19 +1,21 @@
 # tern 0.9.10.9012
 
-### Miscellaneous
-* Updated `roxygen2` to 8.0.0 and added `@exportS3Method` tags for S3 methods in `decorate_grob.R` and `utils_grid.R`.
-* Converted  `s_surv_time()`to exported functions.
-
 ### Enhancements
 * Added `alternative` argument to `s_coxph_pairwise()` to allow one-sided hypothesis testing. 
 * Added `lr_stat_df` to the parameters return list of `s_coxph_pairwise()`.
 * Added `uncond_exact_diff` method to `estimate_proportion_diff()` for the unconditional exact confidence interval for the difference in proportions by inverting one-sided tail tests over a nuisance parameter.
+* Added `cmh_sato` as option for `test_proportion_diff()` to enable CMH test with Sato variance estimator.
 * Added `range_with_cens_info` statistic to `s_surv_time()`.
 * Added `lsmean_se`, `lsmean_ci`, and `lsmean_diffci` statistics to `s_ancova()`.
 * Added `s_ancova()` to exported functions.
 
 ### Bug Fixes
 * Fixed bug in `prop_diff_cmh()` which previously failed when strata combinations had 0 observations.
+* Fixed bug in `prop_cmh()` where previously the Wilson-Hilferty transformation did not always preserve the direction of the effect for one-sided tests.
+
+### Miscellaneous
+* Updated `roxygen2` to 8.0.0 and added `@exportS3Method` tags for S3 methods in `decorate_grob.R` and `utils_grid.R`.
+* Converted  `s_surv_time()`to exported functions.
 
 # tern 0.9.10
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,18 +4,18 @@
 * Added `alternative` argument to `s_coxph_pairwise()` to allow one-sided hypothesis testing. 
 * Added `lr_stat_df` to the parameters return list of `s_coxph_pairwise()`.
 * Added `uncond_exact_diff` method to `estimate_proportion_diff()` for the unconditional exact confidence interval for the difference in proportions by inverting one-sided tail tests over a nuisance parameter.
-* Added `cmh_sato` as option for `test_proportion_diff()` to enable CMH test with Sato variance estimator.
+* Added `cmh_sato` method to `test_proportion_diff()` for CMH testing with the Sato variance estimator. (#1482)
 * Added `range_with_cens_info` statistic to `s_surv_time()`.
 * Added `lsmean_se`, `lsmean_ci`, and `lsmean_diffci` statistics to `s_ancova()`.
 * Added `s_ancova()` to exported functions.
 
 ### Bug Fixes
 * Fixed bug in `prop_diff_cmh()` which previously failed when strata combinations had 0 observations.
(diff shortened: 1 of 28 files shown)


danielinteractive:

NEWS.md
```diff
@@ -1,19 +1,21 @@
...
 * Added `lsmean_se`, `lsmean_ci`, and `lsmean_diffci` statistics to `s_ancova()`.
 * Added `s_ancova()` to exported functions.
 
 ### Bug Fixes
 * Fixed bug in `prop_diff_cmh()` which previously failed when strata combinations had 0 observations.
+* Fixed bug in `prop_cmh()` where previously the Wilson-Hilferty transformation did not always preserve the direction of the effect for one-sided tests.
+
+### Miscellaneous
+* Updated `roxygen2` to 8.0.0 and added `@exportS3Method` tags for S3 methods in `decorate_grob.R` and `utils_grid.R`.
+* Converted  `s_surv_time()`to exported functions.
```
Melkiades: These are unrelated right?

NEWS.md
```diff
@@ -1,19 +1,21 @@
...
 * Added `lr_stat_df` to the parameters return list of `s_coxph_pairwise()`.
 * Added `uncond_exact_diff` method to `estimate_proportion_diff()` for the unconditional exact confidence interval for the difference in proportions by inverting one-sided tail tests over a nuisance parameter.
+* Added `cmh_sato` as option for `test_proportion_diff()` to enable CMH test with Sato variance estimator.
 * Added `range_with_cens_info` statistic to `s_surv_time()`.
 * Added `lsmean_se`, `lsmean_ci`, and `lsmean_diffci` statistics to `s_ancova()`.
 * Added `s_ancova()` to exported functions.
 
 ### Bug Fixes
 * Fixed bug in `prop_diff_cmh()` which previously failed when strata combinations had 0 observations.
+* Fixed bug in `prop_cmh()` where previously the Wilson-Hilferty transformation did not always preserve the direction of the effect for one-sided tests.
```
Melkiades:
```suggestion
* Fixed one-sided p-values in `prop_cmh()` with Wilson-Hilferty transformation — the sign of the effect was lost, producing incorrect p-values for `alternative = "less"` and `alternative = "greater"`.
```

I think this is a bit more precise, but ok with both

NEWS.md
```diff
@@ -1,19 +1,21 @@
...
 
-### Miscellaneous
-* Updated `roxygen2` to 8.0.0 and added `@exportS3Method` tags for S3 methods in `decorate_grob.R` and `utils_grid.R`.
-* Converted  `s_surv_time()`to exported functions.
-
 ### Enhancements
 * Added `alternative` argument to `s_coxph_pairwise()` to allow one-sided hypothesis testing. 
 * Added `lr_stat_df` to the parameters return list of `s_coxph_pairwise()`.
 * Added `uncond_exact_diff` method to `estimate_proportion_diff()` for the unconditional exact confidence interval for the difference in proportions by inverting one-sided tail tests over a nuisance parameter.
+* Added `cmh_sato` as option for `test_proportion_diff()` to enable CMH test with Sato variance estimator.
```
Melkiades:
```suggestion
* Added `cmh_sato` method to `test_proportion_diff()` for CMH testing with the Sato variance estimator. (#1482)
```
similar but highlights the issue

R/prop_diff_test.R
```diff
@@ -340,22 +347,35 @@ prop_cmh <- function(ary,
...
 
-  cmh_res <- stats::mantelhaen.test(ary, correct = FALSE, alternative = alternative)
-
-  if (transform == "none") {
-    cmh_res$p.value
+  z_stat <- if (diff_se == "standard") {
+    mh_res <- stats::mantelhaen.test(ary, correct = FALSE, alternative = alternative)
+    checkmate::assert_true(mh_res$parameter == 1)
+    sqrt(unname(mh_res$statistic)) * ifelse(unname(mh_res$estimate) < 1, 1, -1)
   } else {
```
Melkiades, line 357: Small thought: the sign here comes from the odds ratio (`mh_res$estimate`), which works because OR and proportion difference always agree in direction. But since we are in the proportion difference world, it might be slightly more natural to derive the sign from the CMH weighted difference directly, e.g.:

```r
cmh <- h_diff_cmh(ary)
sqrt(unname(mh_res$statistic)) * sign(cmh$diff_est)
```

This would also make the `standard` and `sato` paths symmetric in how they get the sign. Not a big deal either way though!

Melkiades (commented):
Beside minor comments great work as always! And great catch on the p-value issue ;)

NEWS.md
danielinteractive: yeah I just moved this section "Miscellaneous" down here to start with the more important "Enhancements"

R/prop_diff_test.R
danielinteractive: yeah good idea, however then we need to calculate `cmh` just to get the sign, which seems a bit expensive. I am adding a comment to highlight this connection

Melkiades: thanks

Melkiades (approved):
Good to go!! Thanks Daniel :)

PR #1487, opened 2026-07-02
feat: add factor_level_method, factor_as_factor, factor_level_last_pattern to df_explicit_na

### LINKED ISSUE #1322: [Feature Request]: <add factor_level_method argument for df_explicit_na function>
### Feature description

Hi @shajoezhu, in data preprocessing, we often use **df_explicit_na** function to convert character variables into factor variables. 
df_explicit_na source code uses **factor** function, factor defaults to **sort(unique(x))** to assign the sorting result of data to **level**. 
We found the following issues:

We note that the results of R's **sort** function for default sorting are not consistent with the results of SAS's **proc sort**.
But the sort function method with **"radix"** and the tidyverse **arrange** function give the same results as SAS.

```
> library(tidyverse)
> data <- data.frame(
+   var = c("Cellulitis","COVID-19","Conjunctivitis","_","-","%")
+ )
> 
> sort(data$var)
[1] "-"              "%"              "_"             
[4] "Cellulitis"     "Conjunctivitis" "COVID-19"      
> sort(unique(data$var), method = "radix")
[1] "%"              "-"              "COVID-19"      
[4] "Cellulitis"     "Conjunctivitis" "_"             
> data %>% arrange(var)
             var
1              %
2              -
3       COVID-19
4     Cellulitis
5 Conjunctivitis
6              _
```

We need to specify the factor level as **sort(unique(x), method = "radix")** to get the same factor level order as the SAS proc sort.

```
> factor(data$var)
[1] Cellulitis     COVID-19       Conjunctivitis _             
[5] -              %             
Levels: - % _ Cellulitis Conjunctivitis COVID-19
> factor(data$var, levels = sort(unique(data$var), method = "radix"))
[1] Cellulitis     COVID-19       Conjunctivitis _             
[5] -              %             
Levels: % - COVID-19 Cellulitis Conjunctivitis _
```
Therefore, the result of **df_explicit_na** function transformation is also inconsistent with SAS.
```
> data1 <- df_explicit_na(data)
> data1$var
[1] Cellulitis     COVID-19       Conjunctivitis _             
[5] -              %             
Levels: - % _ Cellulitis Conjunctivitis COVID-19
```
So we want to add factor_level_method argument for df_explicit_na function:

- When **factor_level_method = "data"**, the factor levels are sorted according to the order in which each value first appears in the data, that is, **unique(x)**.
- When **factor_level_method = "sort_auto" or "default"**, factor's level is **sort(unique(x))**.
- When **factor_level_method = "sort_radix"**, the factor level is **sort(unique(x), method = "radix")**.

Furthermore, can we specify how the level of a specific variable should be set by passing a vector with a name? 
such as: **factor_level_method = c("a" = "data", "b" = "sort_radix")**.

### Code of Conduct

### Contribution Guidelines

### Security Policy

kaipingyang:
## Summary

Adds three new arguments to `df_explicit_na()` to give users control over factor level ordering. Motivated by #1322.

### New arguments

- **`factor_level_method`** (`"sort_auto"` / `"sort_radix"` / `"data"`) — controls how levels are ordered when character or logical columns are converted to factors (or existing factors are re-encoded via `factor_as_factor`):
  - `"sort_auto"` (default): `sort(unique(x))` — locale-aware, identical to the original behaviour. No existing code is affected.
  - `"sort_radix"`: `sort(unique(x), method = "radix")` — byte-order (ASCII) sort, not locale-sensitive. Uppercase letters always sort before lowercase, which matches the behaviour of SAS `PROC SORT` on mixed-case data.
  - `"data"`: `unique(x)` — levels follow first-appearance order in the data.

- **`factor_as_factor`** (`FALSE` by default) — when `TRUE`, existing factor columns are also re-encoded using `factor_level_method`. The default `FALSE` preserves the original behaviour (existing factor levels are kept as-is).

- **`factor_level_last_pattern`** (`NULL` by default) — a regular expression; any levels matching it are moved to the end of the level list, immediately before `na_level`. Useful for consistently placing catch-all categories (e.g. `"Other"`, `"Unknown"`) last. Has no effect when `NULL`.

### Backward compatibility

All three arguments have defaults that exactly reproduce the original function behaviour, so no existing code needs to change.

## Test plan

- [ ] `factor_level_method = "sort_auto"` produces identical results to the unmodified function
- [ ] `factor_level_method = "sort_radix"` sorts uppercase before lowercase on mixed-case data
- [ ] `factor_level_method = "data"` preserves first-appearance order
- [ ] `factor_as_factor = TRUE` re-encodes existing factor levels; `FALSE` leaves them unchanged
- [ ] `factor_level_last_pattern` moves matching levels to end; `NULL` is a no-op
- [ ] `na_level` remains the last level in all cases
- [ ] Input validation errors for invalid argument types

Fixes #1322


shajoezhu:
brilliant! @kaipingyang great to see you here!

kaipingyang:
> brilliant! @kaipingyang great to see you here!

Thanks! Happy to address any review comments.

R/df_explicit_na.R
```diff
@@ -66,17 +88,33 @@
...
+#' # Levels matching "^Other" are placed last (before na_level).
+#' df_explicit_na(my_data, factor_level_last_pattern = "^Other")
+#'
 #' @export
 df_explicit_na <- function(data,
                            omit_columns = NULL,
                            char_as_factor = TRUE,
                            logical_as_factor = FALSE,
-                           na_level = "<Missing>") {
+                           factor_as_factor = FALSE,
```
Melkiades: this shifts `na_level` from position 5 to 6, so positional calls break: `df_explicit_na(d, NULL, TRUE, FALSE, "MISS")` now passes "MISS" to factor_as_factor and errors. can we move the 3 new args after `na_level` to keep positional calls working?

tests/testthat/test-df_explicit_na.R
```diff
@@ -90,3 +90,77 @@ testthat::test_that("df_explicit_na just returns unmodified data if all columns
...
+testthat::test_that("factor_level_method = 'sort_radix' uses byte-order sort (uppercase before lowercase)", {
+  # Mixed-case values: sort_auto is locale-aware (case-insensitive on en_US.UTF-8),
+  # sort_radix is byte-order (uppercase < lowercase, consistent with SAS PROC SORT).
+  my_data <- data.frame(x = c("UNKNOWN", "Unknown", "unknown", NA), stringsAsFactors = FALSE)
+  result_radix <- df_explicit_na(my_data, factor_level_method = "sort_radix")
+  result_auto  <- df_explicit_na(my_data, factor_level_method = "sort_auto")
+  # radix: UNKNOWN < Unknown < unknown (ASCII byte order)
+  testthat::expect_equal(levels(result_radix$x), c("UNKNOWN", "Unknown", "unknown", "<Missing>"))
+  # auto and radix differ on mixed-case data
+  testthat::expect_false(identical(levels(result_radix$x), levels(result_auto$x)))
```
Melkiades: this fails in CI. testthat edition 3 sets `LC_COLLATE=C` (via local_reproducible_output), and under C collate sort_auto is byte-order too, so radix == auto here. passes locally only because of en_US.UTF-8. drop this line, the `expect_equal` above is deterministic in any locale

Melkiades (dismissed):
thanks, this is a nice feature and the defaults do reproduce the old behaviour (checked, output is byte-identical to main across char/logical/factor inputs). a few things before merge:

- the sort_radix test fails under `LC_COLLATE=C`, which is what testthat edition 3 forces via `local_reproducible_output()`. under C collate sort_auto is already byte-order so it equals radix and `expect_false(identical(...))` fails. it only passes locally because your shell is en_US.UTF-8. drop that line and keep the exact-order `expect_equal`, that one holds in any locale
- man/ wasn't re-documented, need to run `devtools::document()` and commit (usage/params are out of sync, R CMD check will complain)
- `factor_as_factor` is inserted before `na_level` so positional calls like `df_explicit_na(d, NULL, TRUE, FALSE, "MISS")` now hit factor_as_factor instead of na_level and error. tern uses na_level by name everywhere so we're fine internally, but this is exported and used a lot in study code. can we move the 3 new args after na_level?

also please strip the auto-generated trailer from the PR description.

R/df_explicit_na.R
shajoezhu: @kaipingyang

kaipingyang:

kaipingyang:
All four points addressed:
- Moved `factor_as_factor`, `factor_level_method`, and `factor_level_last_pattern` after `na_level` to preserve positional call compatibility
- Removed the `expect_false(identical(...))` line that fails under `LC_COLLATE=C`
- Re-ran `devtools::document()` and committed the updated `man/df_explicit_na.Rd`
- Stripped the auto-generated trailer from the PR description

All 32 tests pass locally. Please re-review when you have a chance.

shajoezhu:
@copilot fix spell check

shajoezhu (approved):
hi @kaipingyang , thanks for the changes. i made minor update, ci checks out, and downstream scda.test are all ok. thanks! Thanks @Melkiades for the review

PR #1489, opened 2026-07-02
Clean up (re-document, fix linters, update old magritr to native pipe)

munoztd0:
# Pull Request

Same as for rtables -> https://github.com/insightsengineering/rtables/pull/1099


munoztd0:
image
I believe this is just a false positive because the URL does exist

shajoezhu (approved):
lgtm! Thanks @munoztd0

shajoezhu:
> image I believe this is just a false positive because the URL does exist

yes, i agree.

PR #1503, opened 2026-07-28
Add `exclude_rows` argument to `g_forest()` (Solves #1498).

### LINKED ISSUE #1498: [Feature Request]: Allow excluding rows from the `g_forest()`.
## Feature description

In the current implementation of `g_forest()`, the function attempts to plot all values (except of those in `LabelRow`) from the columns specified by the `col_x` and `col_ci` arguments. However, there are cases where some values are `NULL` and should not be included in the forest plot.

The following artificial example illustrates the issue (please ignore the meaning and logic of the numbers):

```r
library(rtables)

DM2 <- subset(
  DM,
  COUNTRY %in% c("USA", "CAN", "CHN") & ARM %in% c("A: Drug X", "B: Placebo")
)
DM2$diff <- "diff"

cfun <- function(df, labelstr, .spl_context, split_vals) {
  leaf_splc <- .spl_context[nrow(.spl_context), ]
  if (leaf_splc$cur_col_split_val %in% split_vals) {
    rcell(nrow(df), label = leaf_splc$value)
  } else {
    rcell(NULL, label = leaf_splc$value)
  }
}

afun <- function(df, labelstr, .spl_context, split_vals) {
  leaf_splc <- .spl_context[nrow(.spl_context), ]
  if (leaf_splc$cur_col_split_val %in% split_vals) {
    rcell(0.5, label = "MyStat")
  } else {
    rcell(15, label = "MyStat")
  }
}

lyt <- basic_table() |>
  split_cols_by("ARM", split_fun = drop_split_levels) |>
  split_cols_by("diff", nested = FALSE) |>
  split_rows_by("COUNTRY", split_fun = drop_split_levels) |>
  summarize_row_groups(
    cfun = cfun,
    format = NULL,
    extra_args = list(split_vals = c("A: Drug X", "B: Placebo"))
  ) |>
  analyze(
    "BMRKR1",
    afun = afun,
    extra_args = list(split_vals = "diff")
  )

tbl <- build_table(lyt, DM2)

row_paths_summary(tbl)
```

In this example, we would like to plot `col_x = 3`. However,
```r
g_forest(tbl, col_x = 3)
Error in `geom_point()`:
! Problem while setting up geom aesthetics.
ℹ Error occurred in the 5th layer.
Caused by error in `check_aesthetics()`:
! Aesthetics must be either length 1 or the same as the data (6).
✖ Fix the following mappings: `x`.
Run `rlang::last_trace()` to see where the error occurred.
```
It fails because some rows in the 3rd column contain `NULL` values.

When we redefine `cfun` to return `NA` instead of `NULL`, `g_forest(tbl, col_x = 3)` no longer fails:

```r
cfun <- function(df, labelstr, .spl_context, split_vals) {
  leaf_splc <- .spl_context[nrow(.spl_context), ]
  if (leaf_splc$cur_col_split_val %in% split_vals) {
    rcell(nrow(df), label = leaf_splc$value)
  } else {
    rcell(NA, label = leaf_splc$value)
  }
}
```

## Possible solutions:

We see two possible approaches to address this issue.

### 1. Automatically exclude non-plottable values

Before plotting, `g_forest()` could automatically exclude rows where the value in `col_x` is `NULL`, or another non-numeric value (e.g., `""`, `character(0)`).

### 2. Introduce a new argument, e.g. `exclude_rows`

Add a new argument, `exclude_rows = NULL`, that allows users to explicitly specify the indices of rows to be omitted from the forest plot. The row indices could be defined relative to either `row_paths_summary(tbl)` or `as_result_df(tbl)`. The latter would likely be simpler, since `g_forest()` already converts the input table using `as_result_df(tbl)`.

The implementation could be as simple as:

```r
tbl_df <- as_result_df(tbl)

if (!is.null(exclude_rows)) {
  tbl_df <- tbl_df[-exclude_rows, ]
}
```

This second approach is more general, as it gives users a straightforward way to exclude arbitrary rows from the forest plot while preserving the existing behavior for all other rows. It also avoids introducing implicit filtering rules based on the contents of `col_x` / `col_ci`, leaving the decision of which rows to plot entirely under the user's control.

### Code of Conduct

### Contribution Guidelines

### Security Policy

wwojciech:
Fixes #1498

### DIFF
--- a/R/g_forest.R
+++ b/R/g_forest.R
@@ -47,6 +47,14 @@
 #'   is no longer used.
 #' @param newpage `r lifecycle::badge("deprecated")` `g_forest` is now generated as a `ggplot` object. This argument
 #'   is no longer used.
+#' @param exclude_rows (`integerish` or `NULL`)\cr vector of positive row
+#'   indices specifying rows to exclude from the forest plot. Row indices are
+#'   specified relative to [rtables::as_result_df()] applied to `tbl`. Values
+#'   must be between 1 and the number of rows in the result data frame, with no
+#'   missing values. The specified rows are removed before plotting. This can be
+#'   used to omit rows that should not be displayed in the forest plot, such as
+#'   rows containing non-plottable values. Defaults to `NULL`, meaning that all
+#'   rows are considered for plotting.
 #'
 #' @return `ggplot` forest plot and table.
 #'
@@ -77,6 +85,7 @@
 #' tbl <- basic_table() |>
 #'   tabulate_rsp_subgroups(df)
 #' g_forest(tbl)
+#' g_forest(tbl, exclude_rows = 1)
 #'
 #' # Odds ratio only table.
 #'
@@ -177,7 +186,8 @@ g_forest <- function(tbl,
                      as_list = FALSE,
                      gp = lifecycle::deprecated(),
                      draw = lifecycle::deprecated(),
-                     newpage = lifecycle::deprecated()) {
+                     newpage = lifecycle::deprecated(),
+                     exclude_rows = NULL) {
   # Deprecated argument warnings
   if (lifecycle::is_present(width_row_names)) {
     lifecycle::deprecate_warn(
@@ -220,6 +230,10 @@ g_forest <- function(tbl,
   checkmate::assert_number(font_size, lower = 0)
   checkmate::assert_character(col, null.ok = TRUE)
   checkmate::assert_true(is.null(col) | length(col) == 1 | length(col) == nrow(tbl))
+  checkmate::assert_integerish(
+    exclude_rows,
+    lower = 1L, upper = nrow(as_result_df(tbl)), any.missing = FALSE, min.len = 1L, null.ok = TRUE
+  )
 
   # Extract info from table
   mat <- matrix_form(tbl, indent_rownames = TRUE)
@@ -241,6 +255,9 @@ g_forest <- function(tbl,
   }
 
   tbl_df <- as_result_df(tbl)
+  if (!is.null(exclude_rows)) {
+    tbl_df <- tbl_df[-exclude_rows, ]
+  }
   dat_cols <- seq(which(names(tbl_df) == "node_class") + 1, ncol(tbl_df))
   tbl_df <- tbl_df[, c(which(names(tbl_df) == "row_num"), dat_cols)]
   names(tbl_df) <- c("row_num", tbl_stats)
@@ -252,7 +269,9 @@ g_forest <- function(tbl,
     tbl_df[["empty_ci"]] <- rep(list(c(NA_real_, NA_real_)), nrow(tbl_df))
     ci_col <- which(names(tbl_df) == "empty_ci")
   }
-  if (length(tbl_df[, ci_col][[1]]) != 2) stop("CI column must have two elements (lower and upper limits).")
+  if (nrow(tbl_df) >= 1 && length(tbl_df[, ci_col][[1]]) != 2) {
+    stop("CI column must have two elements (lower and upper limits).")
+  }
 
   if (!is.null(col_x)) {
     x_col <- col_x + 1
@@ -278,17 +297,16 @@ g_forest <- function(tbl,
   x_labels <- x_at
 
   # Apply log transformation
-  if (logx) {
+  if (logx && nrow(tbl_df) >= 1) {
     x_t <- log(x)
     lwr_t <- log(lwr)
     upr_t <- log(upr)
-    xlim_t <- log(xlim)
   } else {
     x_t <- x
     lwr_t <- lwr
     upr_t <- upr
-    xlim_t <- xlim
   }
+  xlim_t <- if (logx) log(xlim) else xlim
 
   # Set up plot area
   gg_plt <- ggplot(data = tbl_df) +
--- a/R/g_forest.R
+++ b/R/g_forest.R
@@ -49,12 +49,13 @@
 #'   is no longer used.
 #' @param exclude_rows (`integerish` or `NULL`)\cr vector of positive row
 #'   indices specifying rows to exclude from the forest plot. Row indices are
-#'   specified relative to [rtables::as_result_df()] applied to `tbl`. Values
-#'   must be between 1 and the number of rows in the result data frame, with no
-#'   missing values. The specified rows are removed before plotting. This can be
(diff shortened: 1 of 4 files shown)


wwojciech:
Hi @gmbecker - I created this PR following your advice. However, I noticed that some of the checks are failing during the CLA step.

@gmbecker, @danielinteractive - could you add yourselves as reviewers for this PR?

R/g_forest.R
```diff
@@ -47,6 +47,14 @@
 #'   is no longer used.
 #' @param newpage `r lifecycle::badge("deprecated")` `g_forest` is now generated as a `ggplot` object. This argument
 #'   is no longer used.
+#' @param exclude_rows (`integerish` or `NULL`)\cr vector of positive row
+#'   indices specifying rows to exclude from the forest plot. Row indices are
+#'   specified relative to [rtables::as_result_df()] applied to `tbl`. Values
```
danielinteractive:
```suggestion
#'   specified in relation to the data frame obtained by applying [rtables::as_result_df()] to `tbl`. Values
```

R/g_forest.R
```diff
@@ -47,6 +47,14 @@
...
+#'   must be between 1 and the number of rows in the result data frame, with no
+#'   missing values. The specified rows are removed before plotting. This can be
```
danielinteractive:
```suggestion
#'   must not have any missings. The specified rows are removed before plotting. This can be
```

R/g_forest.R
```diff
@@ -241,6 +255,9 @@ g_forest <- function(tbl,
   }
 
   tbl_df <- as_result_df(tbl)
```
danielinteractive: I would move this further up now above the assertion to avoid duplicate calculation of it

R/g_forest.R
```diff
@@ -278,17 +297,16 @@ g_forest <- function(tbl,
   x_labels <- x_at
 
   # Apply log transformation
-  if (logx) {
+  if (logx && nrow(tbl_df) >= 1) {
```
danielinteractive, line 302: why do we now connect this log condition with the number of entries in `tbl_df`?

R/g_forest.R
```diff
@@ -278,17 +297,16 @@ g_forest <- function(tbl,
...
     lwr_t <- log(lwr)
     upr_t <- log(upr)
-    xlim_t <- log(xlim)
   } else {
     x_t <- x
     lwr_t <- lwr
     upr_t <- upr
-    xlim_t <- xlim
   }
+  xlim_t <- if (logx) log(xlim) else xlim
```
danielinteractive, line 311: why is this one treated separately? if needed, please add a code comment

danielinteractive (commented):
Thanks @wwojciech , please see comments below

R/g_forest.R
wwojciech: > why do we now connect this log condition with the number of entries in `tbl_df`?

It is because `nrow(tbl_df) == 0` can now occur as an edge case (when all rows are excluded), and I wanted `g_forest()` to handle this situation gracefully. In that case, `x`, `lwr`, and `upr` are all `NULL`, causing `log()` to fail. 

I used `nrow(tbl_df) >= 1` because it is concise and covers all cases. Alternatively, we could make the condition more explicit by checking `all(!is.null(x), !is.null(lwr), !is.null(upr))`? But I think that the explicit `NULL` checks make sense only if those objects could legitimately be `NULL` for reasons other than an row-empty `tbl_df`.

R/g_forest.R
wwojciech: It is because `xlim` cannot be `NULL` (see also my response above to your previous question).

R/g_forest.R
wwojciech: I did this intentionally because I wanted the assertions to use only the function arguments. Here, `tbl_df` is derived internally, and I think it is preferable for assertion error messages to refer only to the inputs supplied by the user, making them easier to understand. What do you think?

wwojciech:
Thanks, @danielinteractive - I replied to your questions. Regarding the other suggestion, I will apply it soon, as you suggested.

R/g_forest.R
danielinteractive: Hm I think it is still understandable given the argument documentation. I would weigh the duplicate calculation higher here

danielinteractive (approved):
Looks good to me, thanks @wwojciech !

wwojciech:
> Looks good to me, thanks @wwojciech !

Thank you @danielinteractive . Can you pl. merge it? I have no writing rights.

danielinteractive:
@wwojciech I cannot either. @Melkiades and @shajoezhu , could you help us?

R/g_forest.R
```diff
@@ -240,7 +256,9 @@ g_forest <- function(tbl,
     arms <- NULL
   }
 
-  tbl_df <- as_result_df(tbl)
+  if (!is.null(exclude_rows)) {
+    tbl_df <- tbl_df[-exclude_rows, ]
```
Melkiades, line 260: I do not understand the need of this parameter inside this function when you can add the filtering before the input i.e. doing `tbl_df <- tbl_df[-exclude_rows, ]` just before running g_forest()

The rest is fine, it makes the function more stable ^^

wwojciech: > tbl_df <- tbl_df[-exclude_rows, ]

The issue is that using `g_forest(tbl[-exclude_rows, ])` removes rows from the table itself. The intention is to keep these rows visible in the table while only excluding them from the forest plot.

Melkiades: It makes sense! sorry I had to double check its behavior again ^^ then it is is good to go!

Melkiades (approved):
Lgtm! Thanks @wwojciech. Could you only fix cicd?

wwojciech:
> Lgtm! Thanks @wwojciech. Could you only fix cicd?

Thanks a lot, @Melkiades! Actually, I think the CI/CD is failing because I'm merging from a forked repository. I have no idea how to solve it. Maybe it's because my repository is named `tern_ww` instead of `tern`? But I would expect forked repositories to be allowed to have different names.

Is there anyone who could help me with this? Thanks!

shajoezhu:
hi @wwojciech , can you try rename the branch wwojciech:1498_g_forest_exclude_rows , and raise another PR, and let's see. it feels like an cicd with branch naming issue. thanks

PR #1504, opened 2026-08-08
Added the exclude_rows argument to g_forest().

### LINKED ISSUE #1498: [Feature Request]: Allow excluding rows from the `g_forest()`.

wwojciech:
Fixes #1498 (Replaces #1503 due to CI/CD issues).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,8 @@
 # tern 0.9.11.9000
 
 ### Enhancements
+* Added the `exclude_rows` argument to `g_forest()` to allow excluding selected
+  rows from the forest plot before plotting. (#1498)
 * Added `factor_level_method` argument to `df_explicit_na()` to control factor level ordering
   when converting character or logical columns. Supported methods: `"sort_auto"` (default,
   locale-aware, preserves original behavior), `"sort_radix"` (byte-order / ASCII sort), and
--- a/R/g_forest.R
+++ b/R/g_forest.R
@@ -47,6 +47,15 @@
 #'   is no longer used.
 #' @param newpage `r lifecycle::badge("deprecated")` `g_forest` is now generated as a `ggplot` object. This argument
 #'   is no longer used.
+#' @param exclude_rows (`integerish` or `NULL`)\cr vector of positive row
+#'   indices specifying rows to exclude from the forest plot. Row indices are
+#'   specified relative to the data frame obtained by applying
+#'   [rtables::as_result_df()] to `tbl`. No elements of `exclude_rows` may be
+#'   missing.
+#'   The specified rows are removed before plotting. This can be used to omit
+#'   rows that should not be displayed in the forest plot, such as rows
+#'   containing non-plottable values. Defaults to `NULL`, meaning that all rows
+#'   are considered for plotting.
 #'
 #' @return `ggplot` forest plot and table.
 #'
@@ -77,6 +86,7 @@
 #' tbl <- basic_table() |>
 #'   tabulate_rsp_subgroups(df)
 #' g_forest(tbl)
+#' g_forest(tbl, exclude_rows = 1)
 #'
 #' # Odds ratio only table.
 #'
@@ -177,7 +187,8 @@ g_forest <- function(tbl,
                      as_list = FALSE,
                      gp = lifecycle::deprecated(),
                      draw = lifecycle::deprecated(),
-                     newpage = lifecycle::deprecated()) {
+                     newpage = lifecycle::deprecated(),
+                     exclude_rows = NULL) {
   # Deprecated argument warnings
   if (lifecycle::is_present(width_row_names)) {
     lifecycle::deprecate_warn(
@@ -220,6 +231,11 @@ g_forest <- function(tbl,
   checkmate::assert_number(font_size, lower = 0)
   checkmate::assert_character(col, null.ok = TRUE)
   checkmate::assert_true(is.null(col) | length(col) == 1 | length(col) == nrow(tbl))
+  tbl_df <- as_result_df(tbl)
+  checkmate::assert_integerish(
+    exclude_rows,
+    lower = 1L, upper = nrow(tbl_df), any.missing = FALSE, min.len = 1L, null.ok = TRUE
+  )
 
   # Extract info from table
   mat <- matrix_form(tbl, indent_rownames = TRUE)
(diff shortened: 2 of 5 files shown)


wwojciech:

wwojciech:
Hi @shajoezhu, @Melkiades,

I created this new PR to replace #1503.

CI/CD is still failing, and I think there are two main reasons for this:

1. The CLA setup has changed with the move to the new `pharmaverse` model.
2. There are still some old URLs pointing to `insightsengineering` instead of the new `pharmaverse` organization.

Would it be possible for you to update the required files/configuration?

I created #1505 for that.

Thanks!

@danielinteractive , @gmbecker - FYI.

shajoezhu:
it is much better than before, we can fix the url in a seperate PR, that's not a blocker for us to merging this, if only cla check failing, we can fix that later. let's see the others

wwojciech:
> it is much better than before, we can fix the url in a seperate PR, that's not a blocker for us to merging this, if only cla check failing, we can fix that later. let's see the others

Thank you @shajoezhu. If the remaining issues are not blockers, could you please approve this PR so that I can merge it?
PR #1503 was already approved by @danielinteractive and @Melkiades.

PS. I noticed two NOTES in the R CMD check:

1. Example timing (`g_forest()`)
The example is just slightly above the 5-second threshold (user = 5.13s, elapsed = 5.143s). Since the added example is a simple one-line call that executes quickly, I suspect the examples were already close to the limit and this pushed it slightly over.
2. checking CRAN incoming feasibility ... NOTE

shajoezhu:
cran check will not pass 

PS. I noticed two NOTES in the R CMD check:

Example timing (g_forest())
The example is just slightly above the 5-second threshold (user = 5.13s, elapsed = 5.143s). Since the added example is a simple one-line call that executes quickly, I suspect the examples were already close to the limit and this pushed it slightly over.
checking CRAN incoming feasibility ... NOTE
This appears to be a separate CRAN incoming-check NOTE.

can you fix them

wwojciech:
> cran check will not pass
> 
> PS. I noticed two NOTES in the R CMD check:
> 
> Example timing (g_forest()) The example is just slightly above the 5-second threshold (user = 5.13s, elapsed = 5.143s). Since the added example is a simple one-line call that executes quickly, I suspect the examples were already close to the limit and this pushed it slightly over. checking CRAN incoming feasibility ... NOTE This appears to be a separate CRAN incoming-check NOTE.
> 
> can you fix them

I don’t think I can fix these without revisiting the implementation of the existing functions/examples.

The `g_forest()` timing issue appears to be caused by an example that was already close to CRAN’s 5-second threshold, and the additional (1-line trivial) example pushed it slightly over. Fixing that properly would likely require changing the implementation or restructuring the existing examples, neither of which I originally wrote.

I also profiled the `g_forest()` example to identify the source of the timing issue. The main time-consuming operation is `cowplot::plot_grid()`.

The incoming feasibility NOTE also appears to be separate from the added example and not something I can directly address.

shajoezhu:
skip the example if run on cran

wwojciech:
> skip the example if run on cran

Hi @shajoezhu - done. R CMD checks is passsing now.

shajoezhu (approved):
lgtm!

PR #1507, opened 2026-08-10
Update `g_forest()` to support `col_x` and `col_ci` in the same column.

### LINKED ISSUE #1499: [Feature Request]: `g_forest()` to support point estimates and confidence intervals stored in a single column
## Feature description

Currently, `g_forest()` requires the point estimate and confidence interval to be provided in separate columns:

- the column indicated by `col_x` must contain a scalar value representing the point estimate;
- the column indicated by `col_ci` must contain a pair of numbers representing the lower and upper confidence interval limits.

However, some tables store the point estimate and confidence interval limits together in a single column rather than in separate columns.

For example:

```r
library(tern)

d <- formatters::ex_adlb
d$ABLFL_logical <- d$ABLFL == "Y"

results <- basic_table() |>
    split_cols_by("ARM", split_fun = keep_split_levels("A: Drug X")) |>
    split_rows_by("AVISIT", split_fun = keep_split_levels(c(
        "BASELINE",
        "WEEK 1 DAY 8",
        "WEEK 2 DAY 15"
    ))) |>
    summarize_change(
        "CHG",
        variables = list(value = "AVAL", baseline_flag = "ABLFL_logical"),
        .stats = "mean_ci_3d"
    ) |>
    build_table(d)

results
                        A: Drug X      
———————————————————————————————————————
BASELINE                               
  Mean (95% CI)   49.48 (48.66 - 50.31)
WEEK 1 DAY 8                           
  Mean (95% CI)    0.47 (-0.68 - 1.63) 
WEEK 2 DAY 15                          
  Mean (95% CI)    0.48 (-0.70 - 1.66) 
```

## Requested enhancement

Extend `g_forest()` to support tables with a column containing all three values in the following format
```
est (ci_lwr, ci_upr).
```
(or some variation of this, that is: `est (ci_lwr - ci_upr)`, `est (ci_lwr; ci_upr)`, ....

`g_forest()` should be able to extract these three components internally and use them to generate the forest plot.

### Code of Conduct

### Contribution Guidelines

### Security Policy

wwojciech:
Fixes #1499

### DIFF
--- a/R/g_forest.R
+++ b/R/g_forest.R
@@ -334,6 +334,8 @@ g_forest <- function(tbl,
   if (is.null(col_ci)) {
     lwr <- upr <- lwr_t <- upr_t <- rep(NA_real_, nrow(tbl_df))
   } else {
+    checkmate::assert_matrix(x_ci, min.cols = 2, null.ok = TRUE)
+    checkmate::assert_matrix(x_ci_t, min.cols = 2, null.ok = TRUE)
     lwr <- x_ci[, ncol(x_ci) - 1L]
     upr <- x_ci[, ncol(x_ci)]
     lwr_t <- x_ci_t[, ncol(x_ci_t) - 1L]
--- a/R/g_forest.R
+++ b/R/g_forest.R
@@ -280,12 +280,17 @@ g_forest <- function(tbl,
   tbl_df <- tbl_df[, dat_cols, drop = FALSE]
   names(tbl_df) <- make.unique(mat_strings[nlines_hdr, -1])
 
-  # Check table data columns
-  if (!is.null(col_ci)) {
-    if (nrow(tbl_df) >= 1 && length(tbl_df[, col_ci][[1]]) <= 1) {
+  # Check CI column.
+  if (nrow(tbl_df) >= 1 && !is.null(col_ci)) {
+    ci_len <- length(tbl_df[, col_ci][[1]])
+    if (ci_len < 2) {
       stop("CI column must have at least two elements (lower and upper limits).")
     }
+    if (!is.null(col_x) && col_x == col_ci && ci_len != 3) {
+      stop("x / CI column must have three elements (point estimate, lower and upper limits).")
+    }
   }
+
   # col_symbol_size
   sym_size <- if (!is.null(col_symbol_size)) {
     unlist(tbl_df[, col_symbol_size])
--- a/R/g_forest.R
+++ b/R/g_forest.R
@@ -317,7 +317,7 @@ g_forest <- function(tbl,
   }
 
   # Extract vectors: x, lwr, and upr, and their log transformations.
-  # Note that x, lwr, and upr are NULL when x_ci is NULL.
+  # Note that when x_ci is NULL, x_ci[, i] is NULL for any index i.
   # x
   if (is.null(col_x)) {
     x <- x_t <- rep(NA_real_, nrow(tbl_df))
--- a/R/g_forest.R
+++ b/R/g_forest.R
@@ -16,8 +16,12 @@
 #' @param col_ci (`integer(1)` or `NULL`)\cr column index with confidence intervals.
 #'   By default tries to get this from `tbl` attribute `col_ci`, otherwise needs
 #'   to be manually specified. If `NULL`, lines will be excluded from forest plot.
+#'
 #'   The estimator and confidence interval can be stored in the same column.
-#'   In this case, `col_x` and `col_ci` must be the same.
+#'   In this case, `col_x` and `col_ci` must be the same, and the values in each
+#'   row of the column indicated by `col_x`/`col_ci` must be ordered as the
+#'   point estimate, lower confidence interval bound, and upper confidence
+#'   interval bound, respectively.
 #' @param vline (`numeric(1)` or `NULL`)\cr x coordinate for vertical line, if `NULL` then the line is omitted.
 #' @param forest_header (`character(2)`)\cr text displayed to the left and right of `vline`, respectively.
 #'   If `vline = NULL` then `forest_header` is not printed. By default tries to get this from `tbl` attribute
@@ -255,7 +259,6 @@ g_forest <- function(tbl,
   mat_strings <- formatters::mf_strings(mat)
   nlines_hdr <- formatters::mf_nlheader(mat)
   nrows_body <- nrow(mat_strings) - nlines_hdr
-  tbl_stats <- make.unique(mat_strings[nlines_hdr, -1])
 
   # Generate and modify table as ggplot object
   gg_table <- rtable2gg(tbl, fontsize = font_size, colwidths = width_columns, lbl_col_padding = lbl_col_padding) +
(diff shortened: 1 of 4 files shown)


wwojciech:
Hi @danielinteractive , @gmbecker - Can you pl. review this new PR? Thank you!

R/g_forest.R
```diff
@@ -289,10 +301,23 @@ g_forest <- function(tbl,
...
+    x_ci_list <- lapply(x_ci_df, function(col) {
+      matrix(
+        unlist(col),
+        nrow = nrow(tbl_df),
+        byrow = length(col[[1]]) != 1L
+      )
+    })
+    do.call(cbind, x_ci_list)
+  } else {
+    NULL
```
danielinteractive, line 311: in the case that NULL is used here, then the below indexing in lines 318-320 is not meaningful?

tests/testthat/test-g_forest.R
```diff
@@ -148,6 +148,20 @@ testthat::test_that("g_forest works when all rows are excluded", {
...
+    rrow("row 2", rcell(c(11, 7, 13), format = "xx. (xx. - xx.)"))
+  )
+
+  testthat::expect_silent(
+    p <- g_forest(tbl, col_x = 1, col_ci = 1, vline = 10, xlim = c(5, 15), logx = FALSE)
+  )
+
+  expect_snapshot_ggplot("g_forest_the_same_x_ci", p, width = 15, height = 3)
+})
+
```
danielinteractive, line 290: is there a test for the NULL situation I commented above already?

danielinteractive (commented):
Thanks @wwojciech please see my question below

R/g_forest.R
wwojciech: The idea is to have `x`, `lwr`, and `upr` set to `NULL` when `x_ci = NULL`.

tests/testthat/test-g_forest.R
wwojciech: Yes, this is the test: "g_forest works when all rows are excluded".

R/g_forest.R
danielinteractive: ok then maybe add a comment in line 317 accordingly.

wwojciech:
Thanks @danielinteractive , please see the update code.

danielinteractive (approved):

wwojciech:
Hi @danielinteractive - I had to update the code to ensure backward compatibility, specifically to allow `col_x` or `col_ci` to be `NULL`, hiding `x` or `ci` when the point estimate and CI share a column.

If you have a chance, could you please take a look at the updated code?

R/g_forest.R
```diff
@@ -262,57 +275,73 @@ g_forest <- function(tbl,
...
+    x_t <- x_ci_t[, 1L]
   }
-  if (!is.null(col_symbol_size)) {
-    sym_size <- unlist(tbl_df[, col_symbol_size + 1])
+  # ci
+  if (is.null(col_ci)) {
+    lwr <- upr <- lwr_t <- upr_t <- rep(NA_real_, nrow(tbl_df))
   } else {
-    sym_size <- rep(1, nrow(tbl_df))
+    lwr <- x_ci[, ncol(x_ci) - 1]
```
danielinteractive: are we sure here that `ncol(x_ci)` is >= 2?

R/g_forest.R
```diff
@@ -262,57 +275,73 @@ g_forest <- function(tbl,
...
+    upr <- x_ci[, ncol(x_ci)]
+    lwr_t <- x_ci_t[, ncol(x_ci_t) - 1]
```
danielinteractive: same question here

R/g_forest.R
wwojciech: Yes, we are. If `col_ci` is not `NULL` (and `nrow(tbl_df) >= 1`), lines 284--288 ensure that the CI column has at least two elements. Consequently, lines 301--308 ensure that `x_ci` has at least two columns.
If `nrow(tbl_df) == 0`, then `x_ci` is `NULL`, so `x_ci[, i]` is `NULL` for any index `i`.

R/g_forest.R
wwojciech: Same answer: Yes, we are (`x_ci_t` is derived directly from `x_ci`).

R/g_forest.R
danielinteractive: ok I think good to add an assertion here and a corresponding comment (helps with maintenance)

R/g_forest.R
```diff
@@ -262,57 +275,73 @@ g_forest <- function(tbl,
...
-    ci_col <- which(names(tbl_df) == "empty_ci")
+    rep(1, nrow(tbl_df))
   }
-  if (nrow(tbl_df) >= 1 && length(tbl_df[, ci_col][[1]]) != 2) {
-    stop("CI column must have two elements (lower and upper limits).")
+
+  x_ci <- if (nrow(tbl_df) >= 1) {
+    tbl_x_ci <- tbl_df[, unique(c(col_x, col_ci)), drop = FALSE]
+    x_ci_list <- lapply(tbl_x_ci, function(col) {
+      byrow <- length(col[[1]]) != 1L
```
Melkiades: this reads only the first row to pick fill direction. Fine since rtables cols are homogeneous, but a one-line comment saying that would help. non-blocking.

Melkiades (commented):
All in all it looks good to me!! Thanks for the addition @wwojciech. Once Daniel approves we can proceed

R/g_forest.R
wwojciech: Hi @danielinteractive , sure, do you mean the comment like that: 
```r
# At this point, `x_ci` has at least two columns whenever `col_ci` is not NULL and `nrow(tbl_df) >= 1`.
# If `nrow(tbl_df) == 0`, `x_ci` is `NULL`, so `x_ci[, i]` is `NULL` for any i.
```
and then the assertion:
```r
# We essentially need to check the number of columns in `x_ci` (if it is not NULL).
checkmate::assert_matrix(x_ci, min.cols = 2, null.ok = TRUE)
```
?
Isn't redundant - the comment and the assertion?

wwojciech: Okay, I have added a few proposed changes.

danielinteractive (commented):
I would still prefer to have the assertion as well. Reason: If this code is changed in the future ... then this logic might be broken. And you will not see it easily because the code will still run.

wwojciech:
> I would still prefer to have the assertion as well. Reason: If this code is changed in the future ... then this logic might be broken. And you will not see it easily because the code will still run.

Sure, thanks, @danielinteractive. Would the below assertion be sufficient for this purpose?
```r
# We essentially need to check the number of columns in `x_ci` (if it is not NULL).
checkmate::assert_matrix(x_ci, min.cols = 2, null.ok = TRUE)
```

danielinteractive (approved):

wwojciech:
Thank you @danielinteractive.
@Melkiades - can you pl. approve it, (Daniel has just approved it). Thanks!

wwojciech:
Hi @Melkiades - can you pl. have a look at this PR, Thanks!

Melkiades:
> Hi @Melkiades - can you pl. have a look at this PR, Thanks!

Good to go for me! But cannot merge it or approve it... @shajoezhu could you take a look into this?

Melkiades (approved):

wwojciech:
@shajoezhu - can you pl. marge this? Thanks!

Melkiades (approved):

danielinteractive:
@wwojciech could it be that you need to write the CLA comment?

PR #1513, opened 2026-08-27
Add Mantel-Fleiss criterion function mantel_fleiss_crit().

### LINKED ISSUE #1512: [Feature Request]: Implement the Mantel-Fleiss Criterion
### Feature description

https://documentation.sas.com/doc/en/pgmsascdc/v_078/statug/statug_freq_details92.htm

Image

### Code of Conduct

### Contribution Guidelines

### Security Policy

wwojciech:
Fixes #1512


R/prop_diff_test.R
```diff
@@ -481,3 +481,141 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
...
+  MF <- min( # nolintr
+    sum(m_11h) - sum(n_11h_L),
+    sum(n_11h_U) - sum(m_11h)
+  )
+
+  crit <- quote(MF >= 5)
+  is_satisfied <- eval(crit)
+  if (details) {
+    attr(is_satisfied, "value") <- MF
+    attr(is_satisfied, "criterion") <- deparse(crit)
```
danielinteractive: do we need this? it is specified in the function documentation?

R/prop_diff_test.R
```diff
@@ -481,3 +481,141 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
...
+  n_h <- tbl_mrgn["Sum", "Sum", ] # nolintr
+
+  # Expected value of n_11h under the hypothesis of no association
+  # between group and response within stratum h.
+  m_11h <- (n_1.h * n_.1h) / n_h
+  # Lower (L) and upper (U) bounds for n_11h given the marginal totals.
+  n_11h_L <- pmax(0L, n_1.h - n_.2h) # nolintr
+  n_11h_U <- pmin(n_.1h, n_1.h) # nolintr
+
+  MF <- min( # nolintr
```
danielinteractive: I would recommend to reduce the use of math friendly but style unfriendly variable names - then you don't have to use nolintr so often :)

R/prop_diff_test.R
```diff
@@ -481,3 +481,141 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
...
+#' mantel_fleiss_crit(grp, rsp, strata)
+#' mantel_fleiss_crit(grp, rsp, strata, details = TRUE)
+#'
+#' @references
+#' Mantel, N., and Fleiss, J. L. (1980).
+#' Minimum Expected Cell Size Requirements for the Mantel-Haenszel
+#' One-Degree-of-Freedom Chi-Square Test and a Related Rapid Procedure.
+#' \emph{American Journal of Epidemiology}, 112(1), 129--134
+#'
+#' SAS Institute Inc. \emph{SAS/STAT User's Guide: The FREQ Procedure}.
```
danielinteractive: I would drop this reference

danielinteractive (commented):
Thanks @wwojciech , please see few comments below - otherwise looks nice. 
I don't think we need `na.rm` argument when the other functions which are used together with this don't have it.

R/prop_diff_test.R
wwojciech: Yes, I described this in the function doc.

I like these extra attributes, and other functions in `{tern}` sometimes add attributes to the return value with miscellaneous data when required.

Sometimes people might be interested in whether the data result in an MF value close to the threshold. We also have `details = FALSE`, so by default this information is not returned.

So, this is not required, but I think it might be useful and does not cost much.

R/prop_diff_test.R
wwojciech: Drop the SAS ref only or all the references?

R/prop_diff_test.R
wwojciech: Sure; I agree that the variable names are not particularly style-friendly. I just think that, for readability, it is more beneficial to disable the linter on a few lines and keep the variable names matching the mathematical notation/formula exactly, rather than using alternative names.

I’m not insisting on this, though. If you have a strong preference for changing the variable names, I’m happy to do that ( I have just changed `L` to `lwr` and `U` to `upr` ).

R/prop_diff_test.R
danielinteractive: but what does `deparse(crit)` return? Isn't it just returning the expression (MF > 5)? which would be known from the doc already

R/prop_diff_test.R
danielinteractive: I understand your motivation, however other functions in this package also don't use any `.`s for math in the variable names... so I think best would be to stick with that pattern from the package.

R/prop_diff_test.R
wwojciech: Yes, I used `quote()` and `deparse()` mainly to avoid a programming error where the actual condition being evaluated could accidentally differ from the condition stored in the attributes. This way, the criterion shown in the output is directly derived from the condition that is actually evaluated. :)

danielinteractive: yeah but this information is just what the function is doing. Therefore I don't see the point of including it as an attribute.

danielinteractive (approved):

NEWS.md
danielinteractive, line 1: Can you please separate the previous news (should be under `# tern 0.9.11`) see https://cran.r-project.org/web/packages/tern/news/news.html from the upcoming news under development (should be under current heading)

wwojciech: Done, please have a look if this is ok now. Thanks!

R/prop_diff_test.R
```diff
@@ -481,3 +481,137 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
   tbl <- tbl[, c("TRUE", "FALSE")]
   stats::fisher.test(tbl, alternative = alternative)$p.value
 }
+
+#' @title Check the Mantel-Fleiss Criterion
```
Melkiades:
```suggestion
#' Check the Mantel-Fleiss Criterion
```

R/prop_diff_test.R
```diff
@@ -481,3 +481,137 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
...
+#' mantel_fleiss_crit(tbl, include_value = TRUE)
+#'
+#' @references
+#' Mantel, N., and Fleiss, J. L. (1980).
+#' Minimum Expected Cell Size Requirements for the Mantel-Haenszel
+#' One-Degree-of-Freedom Chi-Square Test and a Related Rapid Procedure.
+#' \emph{American Journal of Epidemiology}, 112(1), 129--134.
+#'
+#' @export
+mantel_fleiss_crit <- function(tbl, include_value = FALSE) {
```
Melkiades: Apologies if I am missing something but I do not understand if this is used in prop diff tests? or is it a utility function? If it needs to be used specifically for a table and connected to the rtables/tern machinery please add it correctly to the lists of options for the interested parameters or this addition will stay unknown to users. If it is not to be used here please create a new file with this and add examples with tern code on how it is meant to be used ^^

Melkiades: Is it already connected to the main code?

wwojciech: The purpose of the function is to determine whether the chi-square approximation for the Mantel-Haenszel statistic is appropriate for stratified 2x2 tables. If the Mantel-Fleiss criterion is met, the CMH stratified approach can be used for estimation/testing; otherwise, Fisher's exact test can be used.
See e.g. the SAS documentation: https://documentation.sas.com/doc/en/pgmsascdc/v_078/statug/statug_freq_details92.htm

Yes, it can be considered connected to the main code in the sense that `{tern}` already has the proportion difference testing and CI estimation functionality, and the Mantel-Fleiss criterion is relevant to both of these paths.

So my intention is for this function to support the existing prop-diff functionality in `{tern}`, rather than to introduce a separate user-facing test. 

BTW: Maybe we can place it in `prop_diff.R` then?

Melkiades: Yes it should be close to where it is used in tern code, and that is the issue, I do not see where it is used. Also tests never mention this function into any tern mechanism

Melkiades: and examples

R/prop_diff_test.R
```diff
@@ -481,3 +481,137 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
...
```
Melkiades: This is the core issue. Nothing in tern calls this function, and the tests do not connect it to any tern mechanism either. If it is meant to inform the CMH vs Fisher choice, wire it into s_test_proportion_diff(). If it is a standalone utility, add examples that show real tern usage. Right now it is an orphan export.

R/prop_diff_test.R
```diff
@@ -481,3 +481,137 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
   tbl <- tbl[, c("TRUE", "FALSE")]
   stats::fisher.test(tbl, alternative = alternative)$p.value
 }
+
+#' Check the Mantel-Fleiss Criterion
+#'
+#' @description `r lifecycle::badge("stable")`
```
Melkiades: New export marked stable, should be experimental.

R/prop_diff_test.R
```diff
@@ -481,3 +481,137 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
...
+#'   Whether to include the calculated Mantel-Fleiss statistic as an attribute
+#'   of the result.
+#'
+#' @return A logical value indicating whether the Mantel-Fleiss criterion
+#' is satisfied. If `include_value = TRUE`, the result also contains a
+#' value attribute with the calculated Mantel-Fleiss statistic. If there
+#' are no non-empty strata, the result is `NA` and the value attribute is
+#' `NA_real_`.
+#'
+#' @author WW
```
Melkiades: @author is not used anywhere else in tern, git history covers this. Drop it.

R/prop_diff_test.R
```diff
@@ -481,3 +481,137 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
...
+  # Lower and upper bounds for n_11 given the marginal totals (within a given stratum).
+  n_11_lwr <- pmax(0L, n_1dot - n_dot2)
+  n_11_upr <- pmin(n_dot1, n_1dot)
+
+  mf_value <- min(
+    sum(m_11) - sum(n_11_lwr),
+    sum(n_11_upr) - sum(m_11)
+  )
+
+  is_satisfied <- mf_value >= 5
```
Melkiades: Where 5 comes from? Pull it out to a named constant so the criterion is explicit in the code, not just the docs.

Melkiades (changes requested):
The function itself is fine, math checks out and tests are solid. My blocker is the same as before: it is exported but not used anywhere in tern. s_test_proportion_diff() still picks CMH vs Fisher purely from the method arg, it never calls this. Moving it to prop_diff.R does not fix that, the problem is integration not location.

So a decision is needed: is this an internal helper that drives method selection, or a public function users call themselves?

- internal: drop @export, use the h_ prefix, and call it from s_test_proportion_diff(). This also settles the value/crit attribute thread, no need for it if it is internal.
- public: keep it exported but the examples must show the full tern workflow that builds the tbl, not just table() on random data. Otherwise users have no way to know when or how to use it.

Until one of those is done I do not think this should go in.

R/prop_diff_test.R
wwojciech: Threshold `5` corresponds to the criterion described by Mantel and Fleiss (1980). I will parametrize it anyway.

R/prop_diff_test.R
```diff
@@ -563,12 +563,21 @@ prop_fisher <- function(tbl, alternative = c("two.sided", "less", "greater")) {
...
+#' # Examples of use.
+#'
+#' if (is_mf_satisfied) {
+#'   print("CMH")
+#'   prop_diff_cmh(rsp, grp, strata)$prop
+#' } else {
+#'   print("Exact")
+#'   prop_diff_uncond_exact(rsp, grp)$prop
+#' }
+#'
```
Melkiades, line 575: I would add this into the above mentioned prop_diffs documentation and in the main tabulation function example. So users know it exists and what it can be used for (also add a bit of seealso)

Melkiades: otherwise it is still self-standing and unreachable for users coming from the main tabulation functions.

I would also probably add a verb like estimate_mantel_fleiss_crit so it complies with a more general standard for function naming (I know here in tern it is a bit messy but new code can be better!)

Melkiades (changes requested):
Thanks for this. The stats side is clear and I am happy to have the criterion available, the issue is purely how it connects into tern.

Right now mantel_fleiss_crit() is exported but nothing calls it, no table or analysis function uses it, and no test reaches it through any tern mechanism. The documentation describes it in isolation, unconnected to anything in the package.

Two options:
1. Wire it into the prop diff test path and add a test plus examples that show it in the context where it is actually used, with @seealso pointing to those functions.
2. If it is a standalone utility, say so, move it to its own utils file, and document it as such. If it really is meant to stand alone, a short vignette showcasing when and how to use it would make it discoverable.

Naming should state the action too: check_mantel_fleiss() reads better than mantel_fleiss_crit().

I would rather we solve this here than defer to a follow-up issue. We are trying not to leave orphan functions that get forgotten and unmaintained, so let us land it properly connected in this PR.

R/prop_diff_test.R
Melkiades: Consolidated into the main review, the connect-into-tern point is the blocker there. Resolving here.

R/prop_diff_test.R
Melkiades: Covered in the main review under docs and @seealso. Resolving here.

shajoezhu (approved):
hi @wwojciech , thanks for the changes, i think we can merge this for now, can you create an issue to identify the remaining issue/topic that we can improve for this? such as the sfun function

wwojciech:
@shajoezhu We you merge this please?

NEWS.md
```diff
@@ -1,6 +1,10 @@
 # tern 0.9.11.9000
 
 ### Enhancements
+1512_add_Mantel_Fleiss_criterion
```
Melkiades:
```suggestion
```

Melkiades (approved):
Lgtm! I would love having it user-facing but it is a good addition already! Thanks @wwojciech feel free to merge

wwojciech:
> Lgtm! I would love having it user-facing but it is a good addition already! Thanks @wwojciech feel free to merge

Thank you @Melkiades - a user facing examples will be added within #1528  :)
I would love to merge it but have no rights to do so :)

shajoezhu (approved):
lgtm! thanks @wwojciech , automerge enabled

PR #1515, opened 2026-08-29
Add new (helper) functions for proportions that prepares safe 2x2 cont. tables(s).

### LINKED ISSUE #1514: [Feature Request]: Add `safe_2x2_table()` helper
### Feature description

I suggest adding the following small helper function to the package:

```r
#' @param rsp (`logical`)\cr
#'   A logical vector indicating whether each observation is a responder.
#' @param grp (`factor`)\cr
#'   A factor assigning observations to one of two groups (e.g., reference
#'   and treatment). It must have exactly two levels.
#' @param strata (`factor`)\cr
#'   Factor defining the stratification variable. Each unique
#'   stratum defines a separate 2 x 2 contingency table.
safe_2x2_table <- function(rsp, grp, strata) {
  checkmate::assert_logical(rsp, any.missing = FALSE)
  checkmate::assert_factor(grp, len = length(rsp), any.missing = FALSE, n.levels = 2)
  checkmate::assert_factor(strata, len = length(rsp), any.missing = FALSE)

  # Make rsp a factor to handle cases with only TRUE or only FALSE.
  rsp <- factor(rsp, levels = c("TRUE", "FALSE"))

  table(grp, rsp, strata)
}
```

### Motivation

This could be useful in multiple places, including: `mantel_fleiss_crit()`, `prop_cmh()`, `prop_fisher()`, and other tests for proportion-difference functions

Having a shared helper would avoid duplicating the input validation and, importantly, would consistently handle cases where `rsp` contains only `TRUE` or only `FALSE` observations.

### Code of Conduct

### Contribution Guidelines

### Security Policy

wwojciech:
Fixes #1514 

### What was done:

1. Added new functions:
 - `get_complete_cases()` (removes observations containing missing values) as a prerequisite for `h_prepare_2x2_table()`,
 - `assert_proportion_data()` as an assertion used in `h_prepare_2x2_table()`,
 - `assert_stratification_compatibility()` as an assertion used in `s_proportion_diff()` and  `s_test_proportion_diff()`.

2. Updated the following functions:
- `s_proportion_diff()`,
- `s_test_proportion_diff()`,
so that they use the new functions `h_prepare_2x2_table()` and `assert_stratification_compatibility()`,
- `check_diff_prop_ci()` so that is uses new `assert_proportion_data()`.

3.  Resolved issue #1521.


R/prop_diff.R
```diff
@@ -366,6 +366,51 @@ estimate_proportion_diff <- function(lyt,
...
+#'   `rsp`. Missing values are not allowed.
+#' @param strata (`factor` or `NULL`)\cr
+#'   Defines the stratification variable. When supplied, it must have the same
+#'   length as `rsp` and must not contain missing values.
+#'
+#' @return
+#' Invisibly returns `NULL`. An error is raised if any input does not meet
+#' the required conditions.
+#'
+#' @examples
```
danielinteractive: because this is an internal, not exported, function, there should not be any examples (otherwise there will be a check error)

R/prop_diff.R
```diff
@@ -366,6 +366,51 @@ estimate_proportion_diff <- function(lyt,
...
+#'
+#' \dontrun{
+#' # An error is raised when `grp` has only one level.
+#' grp <- factor(rep("X", 6))
+#' assert_prop_data(rsp, grp, strata)
+#' }
+#'
+#' @author WW
+#'
+#' @export
```
danielinteractive: hm this should be internal only I think

R/prop_diff.R
```diff
@@ -394,16 +441,71 @@ check_diff_prop_ci <- function(rsp,
...
+#' `TRUE` and `FALSE`, in that order, including when one outcome is not observed.
+#'
+#' When `strata` is supplied, a 3-dimensional contingency table with dimensions
+#' 2 x 2 x k is returned, where k is the number of levels of strata.
+#' The dimensions correspond to `grp`, `rsp`, and `strata`, respectively.
+#'
+#' @seealso [assert_prop_data()]
+#'
+#' @author WW
+#' @export
```
danielinteractive, line 480: only export if we need it in `junco` I would say

danielinteractive (commented):
Thanks @wwojciech , please see first comments re: exporting

R/prop_diff.R
wwojciech: Yes, we would need this in junco (in custom afuns).

R/prop_diff.R
wwojciech: Can we use `\dontrun{}` instead? If not, then sure - I can remove it.
I think examples are always very illustrative.

R/prop_diff.R
wwojciech: I agree.

R/prop_diff.R
danielinteractive: for developers tests are sufficient

R/prop_diff.R
wwojciech: @danielinteractive - just realized we need `assert_prop_data()` in `{junco}`, so I need to export it.

danielinteractive (approved):

R/prop_diff.R
```diff
@@ -366,6 +366,50 @@ estimate_proportion_diff <- function(lyt,
...
+#' \dontrun{
+#' # An error is raised when `grp` has only one level.
+#' grp <- factor(rep("X", 6))
+#' assert_proportion_data(rsp, grp, strata)
+#' }
+assert_proportion_data <- function(rsp, grp, strata = NULL) {
+  checkmate::assert_logical(rsp, any.missing = FALSE)
+  checkmate::assert_factor(grp, len = length(rsp), any.missing = FALSE, n.levels = 2)
+  checkmate::assert_factor(strata, len = length(rsp), any.missing = FALSE, null.ok = TRUE)
+  invisible()
```
Melkiades: use the default assertion output

R/prop_diff.R
```diff
@@ -366,6 +366,50 @@ estimate_proportion_diff <- function(lyt,
...
+#' @param strata (`factor` or `NULL`)\cr
+#'   Defines the stratification variable. If not `NULL`, it must have the same
+#'   length as `rsp` and must not contain any missing values.
+#'
+#' @return
+#' Invisibly returns `NULL`. An error is raised if any input does not meet
+#' the required conditions.
+#'
+#' @author WW
+#' @export
```
Melkiades: no export. This is an internal check of inputs. please also add it to the other assertions and not here

R/prop_diff.R
```diff
@@ -376,6 +420,8 @@ estimate_proportion_diff <- function(lyt,
 #' @inheritParams prop_diff
 #' @inheritParams prop_diff_wald
 #'
+#' @seealso [assert_proportion_data()]
```
Melkiades: no need, it is not really fundamental info/function

R/prop_diff.R
```diff
@@ -366,6 +366,50 @@ estimate_proportion_diff <- function(lyt,
...
+#' grp <- factor(c(rep("Placebo", 3), rep("X", 3)))
+#' strata <- factor(c("A", "A", "B", "A", "B", "B"))
+#'
+#' assert_proportion_data(rsp, grp, strata)
+#'
+#' \dontrun{
+#' # An error is raised when `grp` has only one level.
+#' grp <- factor(rep("X", 6))
+#' assert_proportion_data(rsp, grp, strata)
+#' }
```
Melkiades: I would add nord

Melkiades: only for devs

R/prop_diff.R
wwojciech: @Melkiades 
We need this function in our package, when writing custom analysis function, can we therefore export it?
`{tern}` already exports other assertion functions, e.g. 
```r
#' @export
assert_proportion_value <- function(x, include_boundaries = FALSE) {
...
```

R/prop_diff.R
wwojciech: I actually think that from the perspective of `check_diff_prop_ci()` and its user this is fundamental... but ok, I can remove it.

R/prop_diff.R
```diff
@@ -394,16 +394,69 @@ check_diff_prop_ci <- function(rsp,
...
+#' prop_fisher(tbl)
+#'
+#' # The FALSE/TRUE levels are retained even when only one outcome is observed.
+#' safe_2x2_table(rep(TRUE, 8), grp)
+#'
+#' # Stratified 2 x 2 tables.
+#' strata <- factor(c(rep("S1", 4), rep("S2", 4)))
+#'
+#' safe_2x2_table(rsp, grp, strata)
+safe_2x2_table <- function(rsp, grp, strata = NULL) {
```
Melkiades: there is nothing already made to do this? also I would call it contingency_table

wwojciech: What do you mean by: "there is nothing already made to do this?"?

This is strictly for 2x2 tables; and contingency table can be 3x4.

R/prop_diff.R
wwojciech: this is the default assertion output.

R/prop_diff.R
Melkiades: **Duplication not consolidated (main point).** This helper reimplements what `s_test_proportion_diff()` already does inline in `R/prop_diff_test.R` (the `factor(rsp, levels = c("TRUE","FALSE"))` + `switch(method, cmh/... = table(grp, rsp, strata), table(grp, rsp))` block). Issue #1514's motivation is to *avoid* that duplication across `s_test_proportion_diff()`, `prop_cmh()`, `prop_fisher()`, `mantel_fleiss_crit()`. As written the logic now lives in two places instead of one. Suggest refactoring `s_test_proportion_diff()` to call `safe_2x2_table()` here, or noting explicitly that wiring-in is a deferred follow-up.

R/prop_diff.R
```diff
@@ -394,16 +394,69 @@ check_diff_prop_ci <- function(rsp,
...
+#'
+#' When strata is `NULL`, a 2 x 2 table is returned, with `grp` defining the rows
+#' and `rsp` defining the columns. The `rsp` dimension always contains the levels
+#' `TRUE` and `FALSE`, in that order, including when one outcome is not observed.
+#'
+#' When `strata` is supplied, a 3-dimensional contingency table with dimensions
+#' 2 x 2 x k is returned, where k is the number of levels of strata.
+#' The dimensions correspond to `grp`, `rsp`, and `strata`, respectively.
+#'
+#' @author WW
```
Melkiades: `@author` isn't used anywhere else in tern, and it leaks into the generated `.Rd`. Git history already covers attribution. I suggest dropping this tag for consistency with the rest of the package.

R/prop_diff.R
```diff
@@ -394,16 +394,69 @@ check_diff_prop_ci <- function(rsp,
...
+  assert_proportion_data(rsp = rsp, grp = grp, strata = strata)
   checkmate::assert_number(conf_level, lower = 0, upper = 1)
   checkmate::assert_flag(correct, null.ok = TRUE)
+  invisible()
+}
 
-  if (!is.null(strata)) {
-    checkmate::assert_factor(strata, len = length(rsp))
-  }
+#' @title Construct 2 x 2 Contingency Tables Safely
```
Melkiades: Minor style: the rest of tern uses a bare title on the first roxygen line rather than an explicit `@title` tag. Suggest `#' Construct 2 x 2 Contingency Tables Safely` to match convention.

R/prop_diff.R
```diff
@@ -394,16 +394,69 @@ check_diff_prop_ci <- function(rsp,
...
+#'
+#' @description `r lifecycle::badge("stable")`
```
Melkiades: New export tagged `stable` . most new tern functions start as `experimental`. Is `stable` intentional here?

R/utils_checkmate.R
```diff
@@ -180,3 +180,39 @@ assert_proportion_value <- function(x, include_boundaries = FALSE) {
...
   }
 }
+
+#' @describeIn assertions
+#' Validates the data required for a proportion analysis, including responder
+#' status, group assignment, and optional stratification.
+#'
+#' @param rsp (`logical`)\cr
+#'   Indicates whether each observation is a responder (`TRUE`) or a
+#'   non-responder (`FALSE)`. Missing values are not allowed.
```
Melkiades: Misplaced backtick: `` (`FALSE)` `` renders as `FALSE)` instead of `FALSE`. Should be `` (`FALSE`) ``. (Same typo appears in the `safe_2x2_table()` roxygen in `R/prop_diff.R`.)

Melkiades (commented):
Thanks for this!!! the helpers are clean and the test coverage is excellent (empty data, single-outcome, unobserved levels, unused strata levels, sparse tables are all covered).

One substantive point plus a few minor issues, left inline. The main thing: issue #1514 frames this as *removing* duplication across `s_test_proportion_diff()`, `prop_cmh()`, `prop_fisher()`, etc., but as it stands `safe_2x2_table()` isn't applied at any of those call sites and the same `factor(rsp, levels=c("TRUE","FALSE"))` + `switch(table(...))` logic is still duplicated inline in `s_test_proportion_diff()`. Could we wire that call site over in this PR?

I also pushed a small fix to `NEWS.md`: the `# tern 0.9.11` heading had landed above the unreleased `g_forest()` entries (#1499/#1498), which moved them into the already-released 0.9.11 section. I moved the heading below them so they stay under `0.9.11.9000`. Feel free to squash/adjust.

R/prop_diff.R
Melkiades: yes I thought there was something default already shipping it but the function is 2 lines in practice so I believe there is nothing that does it as an already defined tool. And also making it more general as I wanted (just to give it more purpose) does not make sense because it is just `table()` anyway.

Given that, I'd go the other way: keep it specific and internal rather than general and exported. It's a tern-domain convenience (responder × group × strata → the 2×2×k shape the prop_* tests need), not a general tabulator. That also resolves the export question in the other thread,I'd make it h_prop_2x2_table() (or h_rsp_2x2_table()) and not export it.

If you'd rather keep it tiny and skip a dedicated function altogether, the alternative is to just inline the one-liner (factor(rsp, levels = c("TRUE","FALSE"))) at the two call sites but a small named helper is fine and more readable, so no strong objection.

R/prop_diff.R
wwojciech: Yes, this was intentional, since the function is quite well tested and isn’t that esoteric/experimental in terms of its purpose or the algorithm used. 
Who is going to replace the experimental tag with stable later? If you have some well-established process for making that change, then I’m happy to change it to experimental. Please let me know what you think.

R/prop_diff.R
wwojciech: I like this tag because it allows for easy and fast identification of the author of the function in case of any questions related to it, especially in a multi-collaborative environment where many people contribute.

However, I can remove it if you prefer.

R/prop_diff.R
wwojciech: I completely agree with you that `safe_2x2_table()` should be used whenever possible, specifically in `s_test_proportion_diff()` and `tern::prop_diff_*.`

I think it would be better to handle this in a separate issue. I’ve created one for this: #1516

wwojciech: Thank you for sharing the additional context. I see your point. However, I think there is also a good argument for making this function external.

Users (including myself and the `{junco}` package) use the functions provided by `{tern}` with all their specifics. In many cases, when writing custom analyses or content functions, before supplying data to `{tern}` functions that require a valid 2x2 table (such as `tern::prop_diff_*`), users need to reliably create and prepare this data. This is where `safe_2x2_table()` comes in.

Users of `{tern}` can use it to prepare the input data for the `{tern}` functions. Given this, I would keep it as an external function, as it provides a clear and convenient helper for reliably constructing the data argument for the `tern::prop_diff_*` functions.

Melkiades (changes requested):
Thanks for this, the direction is useful and I want it in. A few things to sort out here before merge though.

The main point is still open: safe_2x2_table() duplicates the 2x2 logic already inside s_test_proportion_diff() instead of consolidating it, which is the opposite of what #1514 asked for. Please refactor s_test_proportion_diff() to call it so there is a single source.

Naming: safe_2x2_table() is a 2x2 contingency table, so call it what it is and let the name state the action. Something like make_cont_2x2_table() if it builds the table, or an h_ prefix if it stays internal.

Docs need connecting too. Add `@seealso` to the functions that use it, and examples that show it in the context where it is actually called, not just in isolation.

Export: if the assertion is exported put it next to the other assertions and add it consistently rather than in isolation.

I would rather we close these out in this PR than spin off follow-up issues. We are trying to avoid orphan functions that get forgotten and unmaintained, so it is better to land it fully connected now.

R/prop_diff.R
Melkiades: Folded into the main review, covering export and consolidation. Resolving here to keep things in one place.

R/prop_diff.R
Melkiades: Folded into the main review. Resolving here.

R/prop_diff.R
Melkiades: Covered in the main review under naming and consolidation. Resolving here.

R/prop_diff.R
Melkiades: Not a blocker, dropping this one. Resolving.

wwojciech:
Hi @Melkiades 

Thank you for the very nice discussion we had yesterday. Following our agreements, I updated the `s_test_proportion_diff()` function so that it uses `safe_2x2_table()`. I also added more examples for `safe_2x2_table()`.

Please let me know what you think and if there is anything else I could do for this PR.

wwojciech:
Hi @Melkiades,

Thank you for the very nice discussion we had yesterday.

Following our agreements, I replaced `safe_2x2_table()` with `h_prepare_2x2_table()`, which can later be reused in the Mantel-Fleiss criterion use-case examples, as well as for future planned UI unification across the `prop_diff_` stat functions.

I also added `get_complete_cases()` as a prerequisite for `h_prepare_2x2_table()`.

Then, I updated `s_test_proportion_diff()` to use `h_prepare_2x2_table()`.

More extensive unit tests for `h_prepare_2x2_table()` are still TODO.

Please let me know what you think and if there is anything else I could do for this PR. If you have no objections, I will proceed with writing more unit tests for `h_prepare_2x2_table()`.

wwojciech:
Hi @Melkiades,

Thank you for the very good discussion we had.

### Changes applied:

1. Following our agreements, I replaced `safe_2x2_table()` with `h_prepare_2x2_table()`, which can later be reused in the Mantel-Fleiss criterion use-case examples, as well as for future planned UI unification across the `h_prop_diff` functions.

2. I also added `get_complete_cases()` as a prerequisite for `h_prepare_2x2_table()`.

3. Then, I updated `s_test_proportion_diff()` and `s_proportion_diff()` so that they use the new function `h_prepare_2x2_table()`

4. I created issue #1520 - to unify the UI of `h_prop_diff`. I also created issue #1516 to check if there are some other places where `h_prepare_2x2_table()` can potentially be used.

Please let me know what you think and if there is anything else I could do for this PR.

wwojciech:
Hi @Melkiades,

Thank you for the very good discussion we had.

### Changes applied:

1. Following our agreements, I replaced `safe_2x2_table()` with `h_prepare_2x2_table()`, which can later be reused in the Mantel-Fleiss criterion use-case examples, as well as for future planned UI unification across the `h_prop_diff` functions.

2. I also added `get_complete_cases()` as a prerequisite for `h_prepare_2x2_table()`.

3. Then, I updated `s_test_proportion_diff()` and `s_proportion_diff()` so that they use the new function `h_prepare_2x2_table()`

4. I created issue #1520 - to unify the UI of `h_prop_diff`. I also created issue #1516 to check if there are some other places where `h_prepare_2x2_table()` can potentially be used.

Please let me know what you think and if there is anything else I could do for this PR.

R/prop_diff_test.R
```diff
@@ -50,55 +55,49 @@ NULL
...
-      cmh = table(grp, rsp, strata),
-      cmh_sato = table(grp, rsp, strata),
-      cmh_wh = table(grp, rsp, strata),
-      table(grp, rsp)
+    prepared <- h_prepare_2x2_table(
+      df = df, df_ref = .ref_group, var = .var, val = val,
+      strata_vars = strata_vars,
+      complete_cases = TRUE
     )
+    tbl <- prepared$tbl
```
Melkiades, line 93: I would improve variable names here to be more intuitive (tbl -> rsp_tbl?)

R/prop_diff_test.R
```diff
@@ -50,55 +55,49 @@ NULL
...
+    } else {
+      NULL
     }
 
-    tbl <- switch(method,
-      cmh = table(grp, rsp, strata),
-      cmh_sato = table(grp, rsp, strata),
-      cmh_wh = table(grp, rsp, strata),
-      table(grp, rsp)
+    prepared <- h_prepare_2x2_table(
```
Melkiades, line 88: also maybe rsp_tbl_prepared

R/prop_diff_test.R
wwojciech: Sure, I can, but in the rest of the `{tern}` code this variable is like: `tbl` or `t_tbl` (see e.g.: `prop_diff_wald()` - but I am not convinced to `t_tbl` - like what is the meaning of the prefix `t_`?) What do you think?

R/prop_diff_test.R
wwojciech: yeah, but then it becomes very long, especially later, when you extract elements from it: `rsp_tbl_prepared$rsp`. and actually you use it only just below.

Also, `prepared$rsp` looks nice I think - like "prepared response", `prepared$grp` - "prepared group", etc. 

Alternatively, maybe `trgs`, which stands for table, response, group, strata - which is essentially what this function returns. 

What do you think?

R/prop_diff.R
```diff
@@ -78,92 +85,71 @@ NULL
...
                               ...) {
+  checkmate::assert_data_frame(df)
+  checkmate::assert_string(.var)
+  checkmate::assert_subset(.var, colnames(df), empty.ok = FALSE)
+  checkmate::assert_data_frame(.ref_group, null.ok = TRUE)
+  checkmate::assert_flag(.in_ref_col, null.ok = TRUE)
+  checkmate::assert_list(variables, null.ok = TRUE)
+  assert_proportion_value(conf_level)
+  checkmate::assert_character(method)
+  checkmate::assert_character(weights_method)
```
danielinteractive: `match.arg` would be sufficient for this argument

tests/testthat/test-prop_diff.R
```diff
@@ -688,25 +688,73 @@ testthat::test_that("s_proportion_diff works with uncond_exact_diff", {
...
+      .in_ref_col = FALSE,
+      variables = NULL,
+      method = "cmh",
+      val = "Y"
+    ),
+    "strat"
+  )
+})
+
+test_that("s_proportion_diff errors when strata are provided with a non-stratified method", {
```
danielinteractive:
```suggestion
test_that("s_proportion_diff errors when strata are provided with the non-stratified method `uncond_exact_diff`", {
```

NEWS.md
```diff
@@ -1,12 +1,27 @@
...
   stored in a single column. (#1499)
 * Added the `exclude_rows` argument to `g_forest()` to allow excluding selected
   rows from the forest plot before plotting. (#1498)
+  
+### Miscellaneous
+* Added the `val` argument and refactored `s_proportion_diff()` so that it uses
+  the new function `h_prepare_2x2_table()`.
+* Added the `val` argument and refactored `s_test_proportion_diff()` so that it
+  uses the new function `h_prepare_2x2_table()`.
+
```
danielinteractive, line 25: need to add bug fix entry for the weights argument thing

danielinteractive (commented):
Thanks @wwojciech I selectively looked at the weights story here

R/prop_diff.R
wwojciech: Yes, `match.arg()` is already called inside `prop_diff_strat_nc()`. My thought was to briefly validate `weights_method` earlier, at the `s_proportion_diff()` level, since it is just passed down unchanged. That way, an invalid value would be caught closer to the entry point rather than further down the stack. Do you mean that I should add `match.arg(weights_method)` at the `s_proportion_diff()` level?

danielinteractive: Well so in general if you need the argument here, then assert/match it here - in this case match is better imho. If you don't need it here, better to wait for the downstream assert/match call to happen.

wwojciech: Ok, then, since I just pass it to the downstream fun, I will remove the assertion completely.

R/prop_diff_test.R
Melkiades: It is a good programming standard to be longer with variables names. Also it is important to say which object it is (e.g. tbl) 

not a blocker though

R/prop_diff_test.R
Melkiades: lets make the new code better. rsp_tbl is self explanatory and a better option for this variable

R/utils_checkmate.R
```diff
@@ -180,3 +180,84 @@ assert_proportion_value <- function(x, include_boundaries = FALSE) {
...
+#' not use stratification variables.
+#'
+#' @param method (`character(1)`)\cr Specifies the statistical method.
+#' @param stratified_methods (`character`)\cr Names of the methods that require
+#'   stratified data.
+#' @param strata (`character` or `NULL`)\cr Names of the variables defining the
+#'   strata, or `NULL` if no stratification is used.
+#'
+#' @keywords internal
+assert_stratification_compatibility <- function(method, stratified_methods, strata) {
```
Melkiades, line 234: this is a good naming!!! can be shortened but it is much better like this

R/prop_diff_test.R
wwojciech: Makes sense. I will update the code accordingly.

munoztd0:
Created a scda.test branch with this new feature branch to check on snapshot tests -> https://github.com/insightsengineering/scda.test/actions/runs/34983541222

(had to tweak a little bit the version dependencies and fork WW branch to make it work nicely in scda.test but everything else is exactly the same in terms of code)

wwojciech:
> Created a scda.test branch with this new feature branch to check on snapshot tests -> https://github.com/insightsengineering/scda.test/actions/runs/34983541222
> 
> (had to tweak a little bit the version dependencies and fork WW branch to make it work nicely in scda.test but everything else is exactly the same in terms of code)

Thank you very much indeed @munoztd0 for checking this. This is great!

munoztd0 (approved):
Passed all scda.test snapshot tests

shajoezhu:
hi @munoztd0 , can you share the link of the PR that you created for this, so we can track it. thanks

Melkiades:
I am reviewing this. Let me get back at you today @wwojciech @munoztd0

R/prop_diff.R
```diff
@@ -78,92 +85,68 @@ NULL
...
-    grp <- factor(
-      rep(
-        c("ref", "Not-ref"),
-        c(nrow(.ref_group), nrow(df))
+
+  if (is.null(.in_ref_col) || .in_ref_col) {
+    y <- list(diff = numeric(), diff_ci = numeric())
+  } else {
+    checkmate::assert_false(is.null(.ref_group))
+    assert_stratification_compatibility(
```
Melkiades, line 118: This is a behaviour change. On `main`, an unstratified method with `variables = list(strata = "x")` computed the strata and then ignored it, returning unstratified results. Now it errors.

Erroring is the right call, but it needs its own NEWS entry. Same applies to `s_test_proportion_diff()` with `chisq`/`fisher`/`schouten`.

R/prop_diff.R
```diff
@@ -78,92 +85,68 @@ NULL
...
-                              weights_method = "cmh",
+                              weights_method = c("cmh", "wilson_h"),
+                              val = TRUE,
                               ...) {
+  checkmate::assert_data_frame(df)
+  checkmate::assert_string(.var)
+  checkmate::assert_subset(.var, colnames(df), empty.ok = FALSE)
+  checkmate::assert_data_frame(.ref_group, null.ok = TRUE)
+  checkmate::assert_flag(.in_ref_col, null.ok = TRUE)
+  checkmate::assert_list(variables, null.ok = TRUE)
```
Melkiades, line 106: `variables$weights_method` is silently dropped now. On `main`, `s_proportion_diff()` did `weights_method <- variables$weights_method` when non-`NULL`, which was the only path that actually worked before #1521. Anyone passing `variables = list(strata = "x", weights_method = "wilson_h")` now gets CMH weights and different numbers, with nothing raised.

`assert_list(variables, null.ok = TRUE)` accepts any names, so the old spelling passes straight through. Suggest `checkmate::assert_names(names(variables), subset.of = "strata")` so it fails loudly, or keep reading it with a deprecation warning.

R/prop_diff.R
```diff
@@ -394,16 +374,236 @@ check_diff_prop_ci <- function(rsp,
...
+  checkmate::assert_true(
+    checkmate::test_string(val) || checkmate::test_flag(val)
+  )
+  checkmate::assert_subset(var, colnames(df), empty.ok = FALSE)
+  if (!is.null(df_ref)) {
+    checkmate::assert_subset(var, colnames(df_ref), empty.ok = FALSE)
+  }
+  if (!is.null(strata_vars)) {
+    checkmate::assert_subset(strata_vars, colnames(df), empty.ok = FALSE)
+    checkmate::assert_data_frame(df[strata_vars], types = "factor")
```
Melkiades, line 546: This makes factor strata mandatory. `main` used `interaction()`, which accepts character, and ADaM stratification variables are often character. The test in `tests/testthat/test-prop_diff.R` had to drop `stringsAsFactors = FALSE` to keep passing, which is the tell.

`as_factor_keep_attributes()` is already used in `prop_diff_strat_nc()` and would coerce instead of rejecting. If you prefer to keep the restriction, worth saying in NEWS that it is intentional, since it does fix a latent bug: per-group `interaction()` on character columns can produce mismatched level sets between `df` and `df_ref`.

R/prop_diff.R
```diff
@@ -78,92 +85,68 @@ NULL
...
-  y <- list(diff = numeric(), diff_ci = numeric())
-
-  if (!.in_ref_col) {
-    rsp <- c(.ref_group[[.var]], df[[.var]])
-    grp <- factor(
-      rep(
-        c("ref", "Not-ref"),
-        c(nrow(.ref_group), nrow(df))
+
+  if (is.null(.in_ref_col) || .in_ref_col) {
```
Melkiades, line 114: With `.in_ref_col` defaulted to `NULL`, `s_proportion_diff(df, "rsp")` now returns `list(diff = numeric(), diff_ci = numeric())` instead of erroring on the missing argument. A direct caller who forgets `.ref_group` gets an empty result and no signal.

Suggest erroring when `.in_ref_col` and `.ref_group` are both `NULL`, and keeping this branch for `.in_ref_col = TRUE` only.

Related, on line 114: `assert_false(is.null(.ref_group))` reports `Assertion on 'is.null(.ref_group)' failed: Must be FALSE`, which does not name the argument. `checkmate::assert_data_frame(.ref_group)` would, and you already allow `NULL` on line 104.

_pkgdown.yml
```diff
@@ -113,6 +113,8 @@ reference:
       - -h_xticks
       - -prop_diff
       - check_diff_prop_ci
+      - assert_proportion_data
```
Melkiades: `assert_proportion_data` is an alias of the `assertions` topic, which is already listed at line 198 under "Assertion Functions". pkgdown fails with "Topics must be included in the index only once". Worth running `pkgdown::check_pkgdown()` before merge.

`h_prepare_2x2_table` on the next line is its own topic, so that one is fine.

R/utils_checkmate.R
```diff
@@ -180,3 +180,84 @@ assert_proportion_value <- function(x, include_boundaries = FALSE) {
...
+#' status, group assignment, and optional stratification.
+#'
+#' @param rsp (`logical`)\cr
+#'   Indicates whether each observation is a responder (`TRUE`) or a
+#'   non-responder (`FALSE`). Missing values are not allowed.
+#' @param grp (`factor`)\cr
+#'   Assigns each observation to one of two groups, such as a reference and a
+#'   treatment group. Must have exactly two levels and the same length as
+#'   `rsp`. Missing values are not allowed.
+#' @param strata (`factor` or `NULL`)\cr
```
Melkiades, line 195: This `@param strata` is dropped. `assert_stratification_compatibility()` below declares `@param strata` too (line 230), and since both are `@describeIn assertions` they land on the same page, so the generated `man/assertions.Rd` keeps only one `\item{strata}`: the `character` "names of the variables" version. `assert_proportion_data()` takes a factor of stratum values, so its documented type is now wrong.

Renaming the other one to `strata_vars` would fix it and match `h_prepare_2x2_table()`.

Melkiades (changes requested):
@wwojciech This is a clear improvement on what was there. Centralising the 2x2 construction in `h_prepare_2x2_table()` fixes a real latent bug (the old `factor(c(.ref_group[[.var]], df[[.var]]), levels = c("TRUE", "FALSE"))` silently produced all `NA` when `.var` was a factor), adding `cmh_sato` to the strata-required set closes another gap, and the 21 tests on the new helper cover more than the code they replace. No existing snapshot moved, which is good evidence the numbers are preserved.

Requesting changes on a few things that change behaviour silently or will fail CI. None of them are design problems.

1. Unstratified method plus `strata` now errors where it used to be silently ignored. Right call, but it is a behaviour change and needs its own NEWS entry.
2. `variables$weights_method` is silently dropped, so anyone on the old spelling gets different numbers with no error.
3. Strata columns must now be factors. Character strata used to work.
4. `.in_ref_col = NULL` silently returns an empty result instead of erroring.
5. `_pkgdown.yml` lists the `assertions` topic twice, which will fail the pkgdown build.
6. The duplicated `@param strata` drops `assert_proportion_data()`'s own documentation.

Two smaller ones worth folding in while you are here:

* `val` is not a formal of `estimate_proportion_diff()` or `test_proportion_diff()`, so the new feature is only reachable through `...`. Most users never call `s_*()` directly.
* `expect_error(..., "strat")` matches almost any error in this area. `expect_snapshot(..., error = TRUE)` would lock the real message.

Details inline.

R/prop_diff.R
wwojciech: Regarding `variables$weights_method`: this was never documented as a supported way to specify weights for the Newcombe method. Formally, users were therefore not expected to set the weights this way, especially since there is a dedicated `weights_method` argument for this purpose.

The fact that `variables$weights_method` was used in the function body was simply a bug, as described in #1521. Therefore, I think the most reasonable way to handle this change is to add an entry to `NEWS`.

Regarding the assertion, `checkmate::assert_names(names(variables), subset.of = "strata")`: are you sure this is what we want? If so, having variables as a list doesn't make much sense to me.

In any case, I will add the assertion as requested.

R/prop_diff.R
wwojciech: I think that this is wrong suggestion: "Suggest erroring when `.in_ref_col` and `.ref_group` are both NULL, and keeping this branch for `.in_ref_col = TRUE` only."
It should not be an error in this case. A direct caller (`rtables`) who does not specify the value for `.ref_group`, gets an empty result - this is expected behaviour. I coded it now in a more secure way then it was.


`assert_false(is.null(.ref_group))` reports the argument name, see this: `Error: Assertion on 'is.null(.ref_group)' failed: Must be FALSE.`

wwojciech:
> @wwojciech This is a clear improvement on what was there. Centralising the 2x2 construction in `h_prepare_2x2_table()` fixes a real latent bug (the old `factor(c(.ref_group[[.var]], df[[.var]]), levels = c("TRUE", "FALSE"))` silently produced all `NA` when `.var` was a factor), adding `cmh_sato` to the strata-required set closes another gap, and the 21 tests on the new helper cover more than the code they replace. No existing snapshot moved, which is good evidence the numbers are preserved.
> 
> Requesting changes on a few things that change behaviour silently or will fail CI. None of them are design problems.
> 
> 1. Unstratified method plus `strata` now errors where it used to be silently ignored. Right call, but it is a behaviour change and needs its own NEWS entry.
> 2. `variables$weights_method` is silently dropped, so anyone on the old spelling gets different numbers with no error.
> 3. Strata columns must now be factors. Character strata used to work.
> 4. `.in_ref_col = NULL` silently returns an empty result instead of erroring.
> 5. `_pkgdown.yml` lists the `assertions` topic twice, which will fail the pkgdown build.
> 6. The duplicated `@param strata` drops `assert_proportion_data()`'s own documentation.
> 
> Two smaller ones worth folding in while you are here:
> 
> * `val` is not a formal of `estimate_proportion_diff()` or `test_proportion_diff()`, so the new feature is only reachable through `...`. Most users never call `s_*()` directly.
> * `expect_error(..., "strat")` matches almost any error in this area. `expect_snapshot(..., error = TRUE)` would lock the real message.
> 
> Details inline.

Hi @Melkiades ,

Thank you for your review. I’ve implemented all of your suggestions, except for the one regarding `erroring when .in_ref_col and .ref_group are both NULL` (see my comments inline, above).

For the minor 2 comments:

- I think introducing `val` as a formal argument to `estimate_proportion_diff()` and `test_proportion_diff()` (and its `afun`s) would be a relatively bigger change, since the documentation for these functions, as well as for similar functions, currently states that `var` must be logical. Therefore, I think a reasonable compromise is to keep `val` only in the lower-level `sfuns` functions.

- Regarding the error check:
I agree that this is a compromise between being too specific and not specific enough. However, I think matching `"strat"` is a reasonable level of specificity for this test.

Please let me know if there is anything else that needs to be addressed before this PR can be merged.

Many thanks!

R/prop_diff.R
Melkiades: Agreed, it was a bug and the NEWS entry covers it. On the assertion: `subset.of = "strata"` would be my preference over `set_equal`, it keeps the member optional so `variables = list()` still works, and answers your point about the list still making sense. Not blocking, nothing in tern hits it. Same guard would be worth adding to `s_test_proportion_diff()` at some point.

R/prop_diff.R
Melkiades: You are right about the message, it does name the argument. And fair enough on the empty result, dropping the suggestion.

Melkiades (approved):
All six points addressed, thanks for turning it around quickly. Re-ran the affected suites on your branch, all green.

Two non-blocking leftovers noted inline: `subset.of` over `set_equal` for the `variables` guard, plus the same guard in `s_test_proportion_diff()`, and a doc line noting `val` reaches the layout functions through `...`. Happy for those to be a follow-up.

Nice work on `h_prepare_2x2_table()`, it fixes a real latent bug.

R/prop_diff.R
wwojciech: No, `variables = list()` won't work:
```r
> checkmate::assert_names(names(list()), subset.of = "strata")
Error: Assertion on 'names(list())' failed: Must have names.
```

wwojciech:
> All six points addressed, thanks for turning it around quickly. Re-ran the affected suites on your branch, all green.
> 
> Two non-blocking leftovers noted inline: `subset.of` over `set_equal` for the `variables` guard, plus the same guard in `s_test_proportion_diff()`, and a doc line noting `val` reaches the layout functions through `...`. Happy for those to be a follow-up.
> 
> Nice work on `h_prepare_2x2_table()`, it fixes a real latent bug.

Thank you. I commented on why `subset.of` over `set_equal` for the `variables` guard is not a good choice. 
I added guard for `variables` to `s_test_proportion_diff()`. 
Note that `val` reaches stat function not a layout function, for now, I will leave it without any doc line in layout functions about it. 

Thank you for reviewing it. 
@shajoezhu - this PR was approved by @Melkiades , so I guess it can be merged? Many thanks again!

munoztd0:
> hi @munoztd0 , can you share the link of the PR that you created for this, so we can track it. thanks

Sorry I didn't do a PR just ran the actions before.

Here is the clean all green PR -> https://github.com/insightsengineering/scda.test/pull/253

shajoezhu:
> > All six points addressed, thanks for turning it around quickly. Re-ran the affected suites on your branch, all green.
> > Two non-blocking leftovers noted inline: `subset.of` over `set_equal` for the `variables` guard, plus the same guard in `s_test_proportion_diff()`, and a doc line noting `val` reaches the layout functions through `...`. Happy for those to be a follow-up.
> > Nice work on `h_prepare_2x2_table()`, it fixes a real latent bug.
> 
> Thank you. I commented on why `subset.of` over `set_equal` for the `variables` guard is not a good choice. I added guard for `variables` to `s_test_proportion_diff()`. Note that `val` reaches stat function not a layout function, for now, I will leave it without any doc line in layout functions about it.
> 
> Thank you for reviewing it. @shajoezhu - this PR was approved by @Melkiades , so I guess it can be merged? Many thanks again!

that's fine. thanks so much. i added you to the repo

PR #1526, opened 2026-09-12
Added a `method_only` argument to `d_proportion()`, `d_proportion_diff()`,  `d_test_proportion_diff()`.

### LINKED ISSUE #1525: [Feature Request]: Allow proportion description functions to return method label only.
### Feature description

Update the following description functions in the tern package:

- [x] `d_proportion()`,
- [x] `d_proportion_diff()`,
- [x] `d_test_proportion_diff()`.

by adding a  parameter that allows these functions to return only the full method name (label), without the additional any additions (like CI, etc).

This is required for conditional inference methods that may select between different inference methods depending on the specified criteria. The method label needs to remain consistent regardless of which method is ultimately selected, in order to satisfy the `{rtables}` design assumption that row names are common across all columns.

For conditional methods, the desired label can combine the possible methods, for example: "method1_label / method2_label". Creating such a combined label is possible once the functions above support returning the method label only.

### Code of Conduct

### Contribution Guidelines

### Security Policy

wwojciech:
Fixes #1525

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,9 @@
 # tern 0.9.11.9000
 
 ### Enhancements
+* Added a `method_only` argument to `d_proportion()`, `d_proportion_diff()`, and
+  `d_test_proportion_diff()` to allow returning method labels without additional
+  description. (#1525)
 * Updated `g_forest()` to support point estimates and confidence intervals
   stored in a single column. (#1499)
 * Added the `exclude_rows` argument to `g_forest()` to allow excluding selected
--- a/R/prop_diff.R
+++ b/R/prop_diff.R
@@ -415,9 +415,10 @@ check_diff_prop_ci <- function(rsp,
 #'
 #' @inheritParams s_proportion_diff
 #' @param long (`flag`)\cr whether a long (`TRUE`) or a short (`FALSE`, default)
-#'   description is required. Ignored if `method_only = TRUE`.
+#'   description is required.
 #' @param method_only (`flag`)\cr whether to return only the method description,
 #'   without the confidence interval part of the description.
+#'   If `TRUE`, `conf_level` and `long` are ignored.
 #'
 #' @return A `string` describing the analysis.
 #'
--- a/R/prop_diff.R
+++ b/R/prop_diff.R
@@ -457,13 +457,10 @@ d_proportion_diff <- function(conf_level,
   } else {
     ci_label <- f_conf_level(conf_level)
     if (long) {
+      is_cmh_method <- method %in% c("cmh", "cmh_sato", "cmh_mn")
       ci_label <- paste(
         ci_label,
-        ifelse(
-          method %in% c("cmh", "cmh_sato", "cmh_mn"),
-          "for adjusted difference",
-          "for difference"
-        )
+        ifelse(is_cmh_method, "for adjusted difference", "for difference")
       )
     }
     paste0(ci_label, " (", method_label, ")")
--- a/R/prop_diff.R
+++ b/R/prop_diff.R
@@ -414,29 +414,30 @@ check_diff_prop_ci <- function(rsp,
 #' [s_proportion_diff()].
 #'
 #' @inheritParams s_proportion_diff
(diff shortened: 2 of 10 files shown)


R/prop_diff.R
```diff
@@ -414,29 +414,31 @@ check_diff_prop_ci <- function(rsp,
 #' [s_proportion_diff()].
 #'
 #' @inheritParams s_proportion_diff
-#' @param long (`flag`)\cr whether a long (`TRUE`) or a short (`FALSE`, default) description is required.
+#' @param long (`flag`)\cr whether a long (`TRUE`) or a short (`FALSE`, default)
+#'   description is required.
+#' @param method_only (`flag`)\cr whether to return only the method description,
+#'   without the confidence interval part of the description.
+#'   If `TRUE`, `conf_level` and `long` are ignored.
```
danielinteractive: can we inherit this description instead of duplicating it?

R/prop_diff_test.R
```diff
@@ -108,15 +108,25 @@ s_test_proportion_diff <- function(df,
 #' This is an auxiliary function that describes the analysis in `s_test_proportion_diff`.
 #'
 #' @inheritParams s_test_proportion_diff
+#' @param method_only (`flag`)\cr whether to return only the method description,
+#'   including the alternative hypothesis specification, without the "p-value"
+#'   part of the description.
```
danielinteractive, line 113: can we inherit this description instead of duplicating it?

R/prop_diff_test.R
```diff
@@ -125,12 +135,20 @@ d_test_proportion_diff <- function(method, alternative = c("two.sided", "less",
...
     greater = ", 1-sided, direction greater"
   )
-  paste0("p-value (", meth_part, alt_part, ")")
+
+  method_alt_label <- paste0(method_label, alt_label)
+
+  if (method_only) {
+    method_alt_label
+  } else {
+    paste0("p-value (", method_label, alt_label, ")")
```
danielinteractive:
```suggestion
    paste0("p-value (", method_alt_label, ")")
```

NEWS.md
```diff
@@ -1,6 +1,9 @@
 # tern 0.9.11.9000
 
 ### Enhancements
+* Added a `method_only` argument to `d_proportion()`, `d_proportion_diff()`, and
+  `d_test_proportion_diff()` to allow returning method labels without additional
```
danielinteractive:
```suggestion
  `d_test_proportion_diff()` to allow returning method labels without an additional
```

danielinteractive (commented):
Thanks @wwojciech , nice work, please see few minor comments below

R/prop_diff_test.R
wwojciech: This one is actually not duplicated :)

R/prop_diff_test.R
wwojciech: Thank you @danielinteractive 

Actually, in my application, I need the alternative too; namely, I need both the method and the alternative.

I thought (perhaps a bit naively) that the alternative could still be considered part of the method, which would simplify the UI of the function. Otherwise, it becomes a bit more complicated, as I would probably need to introduce another flag to indicate that an alternative is needed.

But maybe that’s actually okay, but then, what could be the names of the flags - `method_only` seems not to be the best choice :)

Maybe, another approach; instead of  `method_only`, lets introduce params like: `add_ci = TRUE`, `add_pval = TRUE`, `add_alternative = FALSE` - would be good? Then, if someone wants to get only method, one needs to set all these arg to `FALSE`.

Does it look better? What do you think?

danielinteractive: Sorry here I am just trying to keep the code DRY - you already defined above `method_alt_label <- paste0(method_label, alt_label)` so I thought you can just use it here? The behavior should stay exactly the same as you had it before.

I am fine with the current argument, its name and how it works. I don't think we need to get more granular.

wwojciech: Ohh, sure, I misunderstood the suggestion. Thanks!

danielinteractive (approved):

PR #1531, opened 2026-09-19
Solves #1530: Rename and update `h_prepare_2x2_table()`

### LINKED ISSUE #1530: Follow-up minor improvements for `h_prepare_2x2_table()`
### Description of planned changes

As a minor follow-up to #1514, I would like to make the following improvements:

- [x] Improve assertions for `val` in `h_prepare_2x2_table()`

   The type of `val` should be consistent with the type of the response column specified by `var`. Specifically:

   - If `df[[var]]` is logical, `val` should be a single logical value.
   - If `df[[var]]` is character or factor, `val` should be a single character value.
   - The same type compatibility should be checked for `df_ref[[var]]`, when `df_ref` is supplied.

   This would provide an earlier and clearer error when an inappropriate value is supplied and avoid relying on implicit coercion or comparisons between incompatible types.

- [x] Rename `h_prepare_2x2_table()` to `h_prepare_rsp_table()`

   The naming issue was already suggested by @Melkiades. After using the function for a while, I think `h_prepare_rsp_table()` is a better and more accurate name.

   The function does not always return a simple 2x2 table: when `strata_vars` are supplied, it returns a `2x2xk` contingency table and also prepares the response, group, and strata vectors.

   Since the function was only introduced recently, I think this is a good opportunity to rename it before the current name becomes more widely established.

### Motivation

This helper function is used in several places throughout the package, so it is important that it is as robust as possible and fails early when invalid inputs are provided, with clear and informative error messages.

wwojciech:
Fixes #1530.


Melkiades (approved):
Lgtm! thanks @wwojciech :)

PR #1533, opened 2026-09-21
adding reports

shajoezhu:
**What changes are proposed in this pull request?**

* Adds a "Validation report" entry to the pkgdown navbar, resolved per documentation version. (#1533)

pkgdown builds one static site per version, so a hardcoded link would freeze at whichever report was current when it was written. `pkgdown/extra.js` reads the version from the navbar and points the link at the newest report at or below it. Doc versions older than the first report drop the item entirely. `_pkgdown.yml` keeps a static href as the fallback for readers with JavaScript disabled.

Only 0.9.11 currently has a report, so 0.9.11 and later resolve to it and earlier doc versions get no navbar item.

`DESCRIPTION` also swaps `RoxygenNote: 8.0.0` for `Config/roxygen2/version: 8.1.0`, which is unrelated to the navbar link.

Same change as pharmaverse/formatters#377 and pharmaverse/rtables#1122.


Melkiades:
@shajoezhu Same three fixes as on pharmaverse/formatters#377 and pharmaverse/rtables#1122, pushed here too, plus a description since the body was the unfilled template.

* `findLink()` scanned every anchor on the page, so a vignette linking the words "validation report" would have been rewritten instead of the navbar item. Scoped to the navbar.
* `closest()` and `remove()` were the only ES6-era DOM calls in an otherwise strict ES5 file. Swapped for equivalents.
* `pickReport()` took the last match, so it depended on `REPORTS` being sorted. It now takes the maximum. The header comment also claimed adding a version needs "no other change", but the static href in `_pkgdown.yml` that no-JS readers get has to move too.

All four copies of `extra.js` are now byte-identical once the package-specific strings are normalised, so they will not drift.

Verified for tern: the 0.9.11 report is a 200, `main` at 0.9.11.9000 resolves to it, and older doc versions correctly drop the item. Note the branch is currently behind `main`, so it will need an update before merge.

One left for you across all four: an rc such as `0.9.11-rc1` resolves to the `0.9.11` report, since `parseInt("rc1")` is `NaN` and becomes 0, so the rc compares equal to the release. Happy to change it if you would rather rc builds got the previous report.

Melkiades (approved):
Approving. The mechanism is sound and I checked it end to end: the report URL is live, the current version resolves to it, older doc versions correctly drop the item, and the live nesttemplate navbar exposes the version exactly where detectVersion() looks. Three small fixes pushed, described above.

